PR #3, opened 2019-02-14
Format the code using rustfmt and add CI

r-darwish:


atroche:
Thanks @r-darwish ! I didn't realise I could use rustfmt in the IntelliJ plugin. I will from now on 👍

PR #7, opened 2019-02-14
Multiple fixes

r-darwish:


src/process.rs
```diff
@@ -144,22 +156,23 @@ impl Process {
...
         let chrome_output_result = wait_for_mut(
             || {
-                let mut buf = [0; 512];
-                let my_stderr = child_process.stderr.as_mut();
+                let my_stderr = BufReader::new(child_process.stderr.as_mut().unwrap());
                 // TODO: actually handle this error
-                let bytes_read = my_stderr.unwrap().read(&mut buf).unwrap();
-                if bytes_read > 0 {
-                    let chrome_output = String::from_utf8_lossy(&buf);
+                for line in my_stderr.lines() {
```
atroche: Much better!

src/transport.rs
```diff
@@ -16,7 +16,20 @@ use crate::waiting_call_registry::WaitingCallRegistry;
 use crate::web_socket_connection::WebSocketConnection;
 use std::sync::mpsc::Receiver;
 
-pub type SessionId = String;
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct SessionId(String);
```
atroche: can you tell me more about why this is preferable to a type alias? (I don't feel strongly either way, I'm just curious)

src/waiting_call_registry.rs
```diff
@@ -18,11 +18,17 @@ impl IdentifiableResponse for Response {
     }
 }
 
-impl WaitingCallRegistry {
-    pub fn new() -> Self {
+impl Default for WaitingCallRegistry {
```
atroche: Similarly here, what's the advantage? (I'm fairly new to Rust!)

atroche:
We can deal with `files` being an array when we come to it :)

src/cdtp/dom.rs
```diff
@@ -76,7 +76,7 @@ pub mod methods {
     #[derive(Serialize, Debug)]
     #[serde(rename_all = "camelCase")]
     pub struct SetFileInputFiles<'a> {
-        pub files: &'a Vec<&'a str>,
```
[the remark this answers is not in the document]
atroche: Oh, my bad, this is actually a slice, not an array, right? And a `&Vec` will deference to one automatically, right?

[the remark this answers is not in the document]
r-darwish: Yeah, just as you can pass `&String` to a function that requires `&str`, so should you be able to pass `&Vec<T>` to a functino that requires `&[T]`, it's just more generic.

src/transport.rs
r-darwish: When you type `pub type Kilometers = u64;` it doesn't make `Kilometers` a unique type. You can pass any `u64` to a function that requires `Kilometers` and if you also have `pub type Meters = u64;` then Rust won't warn you if you add kilometers to meters. 

I found this because one of the function required `&SessionId`, which is basically `&String`. Clippy complained that you should change that to `&str` (Like I changed `&Vec<t>` to `&[T]`), so I figured that it would be best to actually wrap session ID in a type.

Also, when you wrap types like I did it allows you not to derive traits that doesn't make sense. For example, if you have `struct UserId(u64)` then it makes sense to derive `PartialEq`, but doesn't make sense to derive `Add` because adding two user IDs together doesn't make any sense.

src/waiting_call_registry.rs
r-darwish: Not sure if there's a real advantage. It's just something that clippy suggested. I guess it makes sense. If the `new` function doesn't get any parameters then it's basically like saying that the struct implements `Default`

PR #15, opened 2019-02-15
WIP Make default executable Platform independen

lukaslueg:
Here is some code to make the default executable discoverable on multiple platforms. It allows to execute all tests on Linux, which pass (besides the integration tests, which have to go anyway).

`LaunchOptions` inevitably loses the ability to implement `Default`, because we may not be able to (somewhat) safely discover a default path. It is up to the user (and highly platform dependent) what to do then.

### DIFF
--- a/src/process.rs
+++ b/src/process.rs
@@ -27,7 +27,9 @@ enum ChromeLaunchError {
     NoAvailablePorts,
     #[fail(display = "The chosen debugging port is already in use")]
     DebugPortInUse,
-    #[fail(display = "No applicable default launch options, most likely the chrome executable was not found")]
+    #[fail(
+        display = "No applicable default launch options, most likely the chrome executable was not found"
+    )]
     NoDefaultLaunchOptions,
 }
 
--- a/src/process.rs
+++ b/src/process.rs
@@ -27,6 +27,8 @@ enum ChromeLaunchError {
     NoAvailablePorts,
     #[fail(display = "The chosen debugging port is already in use")]
     DebugPortInUse,
+    #[fail(display = "No applicable default launch options, most likely the chrome executable was not found")]
+    NoDefaultLaunchOptions,
 }
 
 struct TemporaryProcess(Child);
@@ -95,9 +97,18 @@ impl LaunchOptions {
         self.port = port;
         self
     }
+
+    pub fn path(mut self, path: std::path::PathBuf) -> Self {
+        self.path = path;
+        self
+    }
 }
 
 impl Process {
+    pub fn default() -> Result<Self, Error> {
+        Self::new(LaunchOptions::default().ok_or(ChromeLaunchError::NoDefaultLaunchOptions)?)
+    }
+
     pub fn new(launch_options: LaunchOptions) -> Result<Self, Error> {
         info!("Trying to start Chrome");
 
--- a/src/process.rs
+++ b/src/process.rs
@@ -39,19 +39,61 @@ impl Drop for TemporaryProcess {
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/process.rs
```diff
@@ -130,7 +171,7 @@ impl Process {
         }
 
         let process = TemporaryProcess(
-            Command::new(launch_options.path)
+            Command::new(&launch_options.path)
```
atroche: Will this panic on systems where the binary doesn't exist? I'm not sure we ever want to let users construct a `LaunchOptions` where `path` is `None`.

src/process.rs
```diff
@@ -39,21 +39,62 @@ impl Drop for TemporaryProcess {
...
+
+    pub fn headless(mut self, headless: bool) -> Self {
+        self.headless = headless;
+        self
+    }
+
+    pub fn port(mut self, port: Option<u16>) -> Self {
+        self.port = port;
+        self
     }
```
atroche:
```suggestion
    }
    
    pub fn path(mut self, path: std::path::PathBuf) -> Self {
        self.path = path;
        self
    }
```

For when users want everything default except for the path (e.g. sometimes I switch between versions of Chrome).

src/process.rs
lukaslueg: `path` can't be `None` here. If you have a `LaunchOption`, you have _some_ `path`. That's why `LaunchOption` no longer implements `Default`: In order to be able to return `None` if there is no default.

atroche: oh, I see, thanks

src/process.rs
```diff
@@ -39,21 +39,67 @@ impl Drop for TemporaryProcess {
...
+    pub fn default_executable() -> Option<std::path::PathBuf> {
+        // TODO BSDs/Unixes are the same?
+        #[cfg(target_os = "linux")]
+        {
+            // TODO Look at $BROWSER and if it points to a chrome binary
+            // $BROWSER may also provide default arguments, which we may
+            // or may not override later on.
+
+            // TODO More paths!?
+            let default_paths = &["/usr/bin/google-chrome-stable", "/usr/bin/chromium-browser"][..];
```
atroche: Not suggesting we do it in this PR, but we could also search PATH.

atroche:
@lukaslueg I'm happy to merge this, just wanted to check with you first because of the WIP in the title

atroche:
(added you as a collaborator, so merge when ready)

src/process.rs
```diff
@@ -39,21 +39,67 @@ impl Drop for TemporaryProcess {
...
+                    return Some(path.into());
+                }
+            }
+        }
+        // TODO Windows
+        // Check HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\App Paths\chrome.exe
+        None
+    }
+
+    pub fn default() -> Option<Self> {
```
atroche: How would you feel about this being a `Result` instead? If someone calls `LaunchOptions::default()` and gets back `None`, they might not understand why, or what they need to do to fix it.

atroche: And it's also let construction be like this:

`Browser::new(LaunchOptions::default()?)?`

Because you can't use the `?` for `Result`s and `Option`s in the same function, right?

lukaslueg: I've added a default constructor to `Browser`, so the `Browser::new(LaunchOptions::default()?)?` can be written as `Browser::default()?`

lukaslueg:
I merge to master now, can't

atroche:
Not quite sure what you mean. Can you rephrase?

But I'll merge anyway :)

PR #16, opened 2019-02-15
Trim dependencies, remove lockfile

lukaslueg:
This removes the lockfile (which is just a hindrance here), removes too-specific version requirements on some dependencies and most importantly massively trims the dependency-tree by making `websockets` not compile the entire tokio-tree (which is unused anyway).


atroche:
Ah, thanks! I was wondering how to get it to leave the `tokio` dependency out.

PR #31, opened 2019-02-16
WIP: Add groundwork for integration tests

lukaslueg:
This adds some basic code to re-add integration tests: We use `tiny_http` to launch a simple http-server and serve static html from there. Tests can easily serve html content and do their worst about it.

Needs expansion.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,9 @@ which = "2.0.1"
 [target.'cfg(windows)'.dependencies]
 winreg = "0.6"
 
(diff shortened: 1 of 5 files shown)


atroche:
I like it, I can imagine using it straight away to write some integration tests. What do you think needs to be expanded before merging?

lukaslueg:
I'll add some more stuff - will take about an hour - and then merge directly to master if thats ok.

atroche:
👍

Issue #34, opened 2019-02-16
Do not expose internal types, modules or methods


r-darwish:
Currently everything in `lib.rs` is set to `pub`. The crate should only export the types necessary for the user.

lukaslueg:
Yes, there is way too many moving parts between `browser`, `process´, `tab` etc.

atroche:
Yeah, I think some of the modules need to be nested, as well.

atroche:
I reckon this is fixed by https://github.com/atroche/rust-headless-chrome/pull/54

PR #40, opened 2019-02-16
 Make most modules not public

atroche:
Also cheekily included a couple of minor cleanups (deleting unused code).


atroche:
Should fix https://github.com/atroche/rust-headless-chrome/issues/34

atroche:
Makes output of rustdoc look like this now:

image

atroche:
Will merge this afternoon unless there are any objections

lukaslueg:
There should be integration tests for all public API - if it vanishes from `pub`, something breaks.

atroche:
Agreed. Which modules do you think need to be part of the public API?

lukaslueg:
I have no opinion on that, I don't know the actual library well enough. Everything that does not _need_ to be public should be private and for everything public, there should be an integration test. That is, the integration tests demonstrate the public API or the lack thereof.

atroche:
closing in favour of https://github.com/atroche/rust-headless-chrome/pull/54

PR #41, opened 2019-02-17
Handle browser / WS conn being closed while using tab 

### LINKED ISSUE #32: Tabs become unresponsive if Browser is dropped
If we have an instance of `Browser`, create a `Tab`, navigate somewhere, `drop()` the `Browser` and continue using the `Tab`, the process starts consuming 100% CPU and becomes unresponsive.

Demonstrated [here](https://github.com/lukaslueg/rust-headless-chrome/blob/d4b5421ae45f41be4aaf53fba80c35367132ca8e/tests/simple.rs#L21-L23) if we `drop()` the `_browser` before calling `wait_for_element()` on the `tab`.

atroche:
Fixes #32

### DIFF
--- a/tests/simple.rs
+++ b/tests/simple.rs
@@ -27,10 +27,8 @@ fn simple() -> Result<(), failure::Error> {
 #[test]
 fn actions_on_tab_wont_hang_after_browser_drops() -> Result<(), failure::Error> {
     logging::enable_logging();
-    let tab = {
-        let (_server, mut _browser, tab) = dumb_server(include_str!("simple.html"));
-        tab
-    };
+    let (_, browser, tab) = dumb_server(include_str!("simple.html"));
+    drop(browser);
     assert_eq!(true, tab.find_element("div#foobar").is_err());
     Ok(())
 }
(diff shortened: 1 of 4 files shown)


tests/simple.rs
```diff
@@ -24,6 +24,17 @@ fn simple() -> Result<(), failure::Error> {
     Ok(())
 }
 
+#[test]
+fn actions_on_tab_wont_hang_after_browser_drops() -> Result<(), failure::Error> {
+    logging::enable_logging();
+    let tab = {
```
lukaslueg: I can't tell if the fix is actually correct or not, I don't know the underlying code well enough. This is the only nitpick, which could be written as

```rust
// Issue 32, dropping Browser while Tab was still in use caused a hang
let (_, browser, tab) = dumb_server(....);
drop(browser);
assert!(tab.find_element("div#foobar").is_err());
```
The explicit `drop` makes it more obvious what the test is about; the drop being implicit in the `let tab = { ... }`-variant.

atroche: Thanks, fixed. I initially tried `browser.drop()` and `Drop::drop(browser)` and they didn't work. I didn't realise there was also `drop`!

lukaslueg:
LGTM, squash and merge

PR #42, opened 2019-02-17
add extension preloading

ravi0213:
Hi, @atroche , could you please review the changes?
I have also added the sample extension and made the changes in integration test.

### DIFF
--- a/src/process.rs
+++ b/src/process.rs
@@ -59,6 +59,10 @@ pub struct LaunchOptions {
     pub headless: bool,
     pub port: Option<u16>,
     pub path: std::path::PathBuf,
+    /// path of the extension,
+    /// in case of multiple extensions, pass a comma separated string of paths,
+    /// for ex. "/path/of/extension/a,/path/of/extension/b"
+    pub load_extension: Option<String>,
 }
 
 impl LaunchOptions {
@@ -101,6 +105,7 @@ impl LaunchOptions {
             headless: true,
             port: None,
             path: Self::default_executable()?,
+            load_extension: None
         })
     }
 
@@ -193,6 +198,13 @@ impl Process {
 
         if launch_options.headless {
             args.extend(&["--headless"]);
+        }
+
+        let mut load_extension_arg:String = String::from("--load-extension=");
+
+        if let Some(load_extension_path) = &launch_options.load_extension {
+            load_extension_arg.push_str(load_extension_path);
+            args.extend(&[load_extension_arg.as_str()]);
         }
 
         let process = TemporaryProcess(
--- a/tests/extension.rs
+++ b/tests/extension.rs
@@ -1,0 +1,17 @@
+use headless_chrome::{browser, logging, process, tab};
+use headless_chrome::LaunchOptions;
+
+#[test]
+fn test_extension() -> Result<(), failure::Error> {
+    let browser = browser::Browser::new(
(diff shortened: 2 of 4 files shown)


src/process.rs
```diff
@@ -129,6 +131,17 @@ impl Process {
             args.extend(&["--headless"]);
         }
 
+        let extension_path = if let Some(load_extension_path) = launch_options.load_extension {
```
atroche: You could have the `format!` and the `args.extend` all within the one `if let Some(load_extension_path) = …` block, if you wanted, and not worry about making the `extension_path` variable.

ravi0213: thank for the feedback, i have made changes as per your feedback :)

lukaslueg:
The integration test is gone? :-)

ravi0213:
> The integration test is gone? :-)

@lukaslueg i think @atroche will add them later.

src/process.rs
```diff
@@ -195,6 +197,13 @@ impl Process {
             args.extend(&["--headless"]);
         }
 
+        let mut load_extensino_arg:String = String::from("--load-extension=");
+
+        if let Some(load_extension_path) = &launch_options.load_extension {
+            load_extensino_arg.push_str(load_extension_path);
```
lukaslueg: This is a typo

src/process.rs
```diff
@@ -59,6 +59,7 @@ pub struct LaunchOptions {
     pub headless: bool,
     pub port: Option<u16>,
     pub path: std::path::PathBuf,
+    pub load_extension: Option<String>,
```
lukaslueg: Isn't it possible to load multiple extensions? Then this would be a `Vec<String>`

ravi0213: Yes, you can load multiple extensions. you need to pass comma separated string of extensions path.

lukaslueg: Could you put that in a docstring?

ravi0213: Sure, will do it. :)

lukaslueg:
The old integration tests have been removed and the new framework is already merged. If you'd like, you could port your original code to whats in `test/simple.rs` and call it `tests/extensions.rs` or something similar.

src/process.rs
atroche: I reckon having this field be a `Vec<String>` and then doing the comma separation (via [`.join()`](https://doc.rust-lang.org/std/slice/trait.SliceConcatExt.html#tymethod.join)) would be pretty neat. And something like `extension_paths` might be a clearer name? But I also don't want to overload you with changes, because what you've already got is good too :)

lukaslueg: If the CLI-parameter takes a comma-separated string, we should do too IMHO. There could be a method `add_extension`, which pushes a comma and the given name.

atroche: Fair enough, I'm cool with that. Looks like Puppeteer itself just lets you pass along arbitrary args to the Chrome binary: https://pptr.dev/#?product=Puppeteer&version=v1.12.2&show=api-working-with-chrome-extensions.

On a side note, it's interesting / worrying that loading extensions apparently won't work in headless mode (found via link above). I wonder how many things there are like that?

atroche:
Hey @ravi0213, thanks for kicking this off! I reckon https://github.com/atroche/rust-headless-chrome/pull/69 should close it out (and it's based on your commit!)

PR #43, opened 2019-02-17
Add Node::find

lukaslueg:
This adds a little comfort to handling `Node`


lukaslueg:
Merging despite CI failure, which is unrelated

Issue #14, opened 2019-02-15
Make methods return &(mut) self where applicable


lukaslueg:
A lot of methods currently return `Result<(), Error>`. More comfort could be provided by returning `Result<&Self, Error>` instead. This causes no change if the `Ok` is not of interest. Yet long chains like

```rust
let tab = browser.wait_for_initial_tab()?;
tab.navigate_to("https://www.wikipedia.org")?;
tab.wait_for_element(r#"input#searchInput"#)?.click()?;
tab.type_str("WebKit")?;
tab.press_key("Enter")?;
```

could be written as

```rust
let tab = browser.wait_for_initial_tab()?;
tab.navigate_to("https://www.wikipedia.org")?
   .wait_for_element(r#"input#searchInput"#)?.click()?
   .type_str("WebKit")?
   .press_key("Enter")?;
```

atroche:
Yeah, I'm into that. Are you also proposing putting the `type_str` and `press_key` methods on `Element` as well?

lukaslueg:
no, sorry, this was a cooked-up example and I did not check if the types actually chain.

lukaslueg:
closed in 6f337722612ccd20646f80a01070d38625b8b009

PR #44, opened 2019-02-17
Allow methods to be chained

lukaslueg:
Some fixes for issue #14 

Some housekeeping on gitignore as well


lukaslueg:
Merging despite CI failure, which is unrelated.

PR #49, opened 2019-02-17
Implement reload

lukaslueg:
The effect of `scriptToEvaluateOnLoad` is unclear to me, I could not get it to do anything...

### DIFF
--- a/src/cdtp/page.rs
+++ b/src/cdtp/page.rs
@@ -80,6 +80,21 @@ pub mod methods {
 
     #[derive(Serialize, Debug)]
     #[serde(rename_all = "camelCase")]
+    pub struct Reload<'a> {
+        pub ignore_cache: bool,
+        #[serde(skip_serializing_if = "Option::is_none")]
+        pub script_to_evaluate: Option<&'a str>,
+    }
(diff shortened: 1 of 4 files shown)


src/cdtp/page.rs
```diff
@@ -78,6 +78,21 @@ pub mod methods {
         type ReturnObject = CaptureScreenshotReturnObject;
     }
 
+    #[derive(Serialize, Debug)]
+    #[serde(rename_all = "camelCase")]
+    pub struct Reload<'a> {
+        pub ignore_cache: bool,
```
atroche: This is an optional param at the protocol level. I'm not sure how important it is to have 1:1 parity, just thought I'd mention it.

atroche:
LGTM

src/cdtp/page.rs
lukaslueg: Yes, I saw that. Yet I'm hesitant to expose any optional parameter that has some default on the protocol level to the API, because every API-call looks like "here are a bunch of parameters and they are all their implicit default, go figure!" I'd strongly favor having less `Option`-type parameters on the public API.

PR #51, opened 2019-02-18
Add "real world" examples

atroche:
These were originally my "integration tests", but they're too network-dependent and flaky for that. I had deleted them, but they're probably worth keeping in `examples/` for now as a source of documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ fn browse_wikipedia() -> Result<(), failure::Error> {
 assert!(browse_wikipedia().is_ok());
 ```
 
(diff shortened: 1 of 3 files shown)


atroche:
Oh, I'm also re-exporting `Tab` from in the lib root (even though I think the `tab` module as a whole should probably not be `pub`), because it's useful to have functions that take / return a tab, e.g.: https://github.com/atroche/rust-headless-chrome/blob/191bd00e8f6cc73dbea9e047fd542ef177168f55/examples/real_world.rs#L202

Issue #19, opened 2019-02-16
Add more of the protocol's Node type's fields to our corresponding struct


atroche:
There are 28 fields listed here: https://chromedevtools.github.io/devtools-protocol/tot/DOM#type-Node

We currently only deserialize 7:

https://github.com/atroche/rust-headless-chrome/blob/dcc5249cf6c890750a901c24b4e2527880b0b1d1/src/cdtp/dom.rs#L9-L18

leshow:
I'm happy to take this issue and the one for deserializing as a HashMap instead of Vec. What should the Map be keyed on? the NodeId?

atroche:
So the `attributes` field is listed as an array of strings representing a map, so `HashMap<String, String>` seems like a good first pass, although we could type alias those strings (or something like https://github.com/atroche/rust-headless-chrome/pull/7#discussion_r257049091 for the attribute name / values).

And, thanks! Yell if you need a hand.

lukaslueg:
Please, lets keep this type-safe and not use a `HashMap`. If the protocol changes and fields come and go, we can always have serde handle optional fields. At some point all the `pub` fields may have to move behind accessors and maybe even `InternalNode`. Having a HashMap here feels like writing Python in Rust.

atroche:
@lukaslueg I think you might have thought I was suggesting that the entire `Node` be represented as  a HashMap, but I was only proposing it for this field in the `Node` struct:

<img width="762" alt="image" src="https://user-images.githubusercontent.com/29839/52928449-cdf36400-3393-11e9-8fc0-6b6e924427ed.png">

Given the format specified in the docs, it seems like a type safe thing to do.

atroche:
(oh, and here's where I proposed it! just realised this issue was never explicitly reference in the above conversation: https://github.com/atroche/rust-headless-chrome/issues/18)

Issue #18, opened 2019-02-16
Deserialize a DOM node's attributes as a HashMap rather than a Vec<String>


atroche:
See:

https://github.com/atroche/rust-headless-chrome/blob/dcc5249cf6c890750a901c24b4e2527880b0b1d1/src/cdtp/dom.rs#L5

and

https://chromedevtools.github.io/devtools-protocol/tot/DOM#type-Node

PR #52, opened 2019-02-18
Convert NodeAttributes to Map, add Node fields

leshow:
closes: #18 
closes: #19 
If it's possible to implement that custom deserialize without an explicit clone on the values I wasn't able to do it. I get a 'cannot move out of borrowed content' borrowck error if I don't clone. Any ideas?

Something I learned: when you do custom deserialize for Option values you need `#[serde(default, ... )]` otherwise serde will panic. I tested this out on a few pages to make sure it could print node attributes properly.

### DIFF
--- a/src/cdtp/dom.rs
+++ b/src/cdtp/dom.rs
@@ -1,8 +1,9 @@
-use serde::Deserialize;
+use serde::{Deserialize, Deserializer};
+use std::collections::HashMap;
 
 pub type NodeId = u16;
 
-pub type NodeAttributes = Vec<String>;
+pub type NodeAttributes = HashMap<String, String>;
 
 #[derive(Deserialize, Debug)]
 #[serde(rename_all = "camelCase")]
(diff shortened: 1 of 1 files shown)


src/cdtp/dom.rs
```diff
@@ -1,19 +1,96 @@
...
+    Close,
+}
+
+fn attribute_deser<'de, D>(d: D) -> Result<Option<NodeAttributes>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let opt: Option<Vec<String>> = Option::deserialize(d)?;
+    Ok(opt.map(|attr| {
+        let mut map = HashMap::new();
```
atroche: What do you think of [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=6322b7504490fb754be4af06ab63a989)?

I'm happy either way, I just wanted to mess around with iterators ;)

atroche:
As for avoiding the explicit clones, I found this way at least:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7313b403324243033dfc8718c55c362f

atroche:
P.S. if you have that test code you mentioned for checking node attributes, paste it here and I'll try to put it with the rest of our tests

src/cdtp/dom.rs
```diff
@@ -1,19 +1,96 @@
...
+    UserAgent,
+    Open,
+    Close,
+}
+
+fn attribute_deser<'de, D>(d: D) -> Result<Option<NodeAttributes>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let opt: Option<Vec<String>> = Option::deserialize(d)?;
```
atroche: What happens if you deserialize into a `Vec<String>` instead of an `Option`?

leshow: We don't want to do that, in that case even with an element with no attributes, you'll end up returning an empty HashMap: `Some({})`

src/cdtp/dom.rs
leshow: Ah, that's the combinator I was looking for. However, I think we can leave it as an explicit loop to avoid the clone?

leshow:
I rebased off the latest master

src/cdtp/dom.rs
atroche: Interesting. Perhaps for elements with no attributes they send back an empty array rather than leaving the field off.

atroche:
Great work.

src/cdtp/dom.rs
leshow: It's a little odd, I get `None` for some elements, and `Some({})` for others. I tested by just running:

```
use headless_chrome::browser::{Browser, LaunchOptions};
use std::error::Error;

fn main() -> Result<(), Box<Error>> {
    let options = LaunchOptions::default().expect("Failed to find chrome");
    let browser = Browser::new(options)?;

    let tab = browser.wait_for_initial_tab()?;

    tab.navigate_to("https://www.fastmail.com/login");
    let element = tab.wait_for_element(r"#header")?;

    println!("{:?}", element.get_description()?);
    Ok(())
}
```

and observing the output:
``` 
cargo run | rg attributes
```

You will see from this that some attributes are None and some are the empty HashMap. I have a suspicion the issue is happening in the lib that's doing the selecting, not in serde.

leshow:
Thank you! I'd be interested in contributing more to this project, if you have anything else you'd like a hand with, feel free to send it my way.

PR #54, opened 2019-02-18
Reorganise modules, and reduce surface area of public API

atroche:
Under these changes, the front page of the docs would look like this:

image

And the file structure would look like this:

image

Let me know what you think! I'll get outstanding PRs merged before merging this one, and then resolve conflicts myself.


src/cdtp/page.rs
```diff
@@ -65,8 +73,8 @@ pub mod methods {
 
     #[derive(Serialize, Debug)]
     #[serde(rename_all = "camelCase")]
-    pub struct CaptureScreenshot<'a> {
-        pub format: Option<&'a str>,
+    pub struct CaptureScreenshot {
+        pub format: Option<super::ScreenshotFormat>,
```
atroche: @lukaslueg how do you feel about having this here instead of in the `tab` module, and relying on `serde` to serialize rather than implementing `AsRef`? The exact representation of the format param seems to me to be a protocol level detail.

lukaslueg: That'd be fine, we can always expand if needed

lukaslueg: I suggest to merge this change, I'll change it further to take a `enum ScreenshotFormat { JPEG(Option<u8>), PNG }` and fix up the protocol API call.

atroche: oh, right, good idea, because it doesn't make much sense to specify quality for png.

PR #56, opened 2019-02-18
Add new_tab method to browser

atroche:
image

### DIFF
--- a/src/browser/mod.rs
+++ b/src/browser/mod.rs
@@ -99,20 +99,20 @@ impl Browser {
 
     /// Create a new tab and return a handle to it.
     ///
-/// ```rust
-/// # use failure::Error;
-/// # fn main() -> Result<(), Error> {
-/// #
-/// # use headless_chrome::{Browser, LaunchOptions};
-/// # let browser = Browser::new(LaunchOptions::default().unwrap())?;
-/// let first_tab = browser.wait_for_initial_tab()?;
-/// let new_tab = browser.new_tab()?;
-/// let num_tabs = browser.get_tabs().lock().unwrap().len();
-/// assert_eq!(2, num_tabs);
-/// #
-/// # Ok(())
-/// # }
-/// ```
+    /// ```rust
+    /// # use failure::Error;
+    /// # fn main() -> Result<(), Error> {
+    /// #
+    /// # use headless_chrome::{Browser, LaunchOptions};
+    /// # let browser = Browser::new(LaunchOptions::default().unwrap())?;
+    /// let first_tab = browser.wait_for_initial_tab()?;
+    /// let new_tab = browser.new_tab()?;
+    /// let num_tabs = browser.get_tabs().lock().unwrap().len();
+    /// assert_eq!(2, num_tabs);
+    /// #
+    /// # Ok(())
+    /// # }
+    /// ```
     ///
     /// Currently does not support creating the tab in a new "browser context", aka an incognito
     /// window.
--- a/src/browser/mod.rs
+++ b/src/browser/mod.rs
@@ -6,7 +6,7 @@ use failure::Error;
 use log::*;
 use serde;
 
-use crate::cdtp::target::methods::SetDiscoverTargets;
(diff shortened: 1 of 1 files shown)


src/browser/mod.rs
```diff
@@ -97,22 +97,49 @@ impl Browser {
...
-    //    }
+    /// Create a new tab and return a handle to it.
+    ///
+    /// ```rust
+    /// # use failure::Error;
+    /// # fn main() -> Result<(), Error> {
+    /// #
+    /// # use headless_chrome::{Browser, LaunchOptions};
+    /// # let browser = Browser::new(LaunchOptions::default().unwrap())?;
+    /// let first_tab = browser.wait_for_initial_tab()?;
```
lukaslueg: What's the significance of waiting for the first new tab here? Do we need one to get a second?

If `wait_for_inititial_tab` does two things, actually waiting for chrome to set itself up and returning the initial tab, I'd suggest we split the API.

src/browser/mod.rs
```diff
@@ -97,22 +97,49 @@ impl Browser {
...
+    /// assert_eq!(2, num_tabs);
+    /// #
+    /// # Ok(())
+    /// # }
+    /// ```
+    ///
+    /// Currently does not support creating the tab in a new "browser context", aka an incognito
+    /// window.
+    pub fn new_tab(&self) -> Result<Arc<Tab>, Error> {
+        let create_target = CreateTarget {
```
lukaslueg: I'd be nice to expose `CreateTarget` to the public in some form, because we might want to control those things

src/browser/mod.rs
atroche: Good point. It's probably better to have `Browser::new` wait for the first tab to be “discovered” (see https://github.com/atroche/rust-headless-chrome/blob/fb80bb14fe44d261a4efdef00ada2ba8dad23fb9/src/browser/mod.rs#L76) before returning. The delay would likely be unnoticeable, and right now there's a brief window in time where you can have a handle to Browser and have `get_tabs()` return a vector of length zero, which is a bit inconvenient.

src/browser/mod.rs
atroche: Yeah, I'm finding that I'm really missing "optional function parameters" as a language feature, or overloaded constructors.

As in, it'd be nice to have a `new_tab` method where you didn't have to create a struct, while also having one where you can specify all the things in `CreateTarget`. And having two seperate functions to represent those things seems cumbersome.

The closest thing I've found is implementing `Default`, but even then…

lukaslueg: This can always be put off for later. I don't think a `pub fn new_tab_with_target()` is not bad at all, something to the tune of

```rust
let new_tab = browser.new_tab_with_target().width(1024).height(768)?;
````

can be constructed if `new_tab_with_target()` puts `&self` into a builder-struct, returns that and the builder-struct returns the actual new tab or the error via the `try` trait.

atroche: Yeah, I'll merge this for now. Also, on a side note, I'm thinking maybe we should rename `Tab` to `Page` to keep it more in line with both Puppeteer and the Chrome DevTools Protocol in general.

When I started the library I wasn't very clear on the exact distinctions between (in CDTP-speak) Target / Page / Frame.

atroche: Related conversation: https://www.reddit.com/r/rust/comments/aefzwk/overloaded_functions_are_now_possible_in_nightly/

atroche: (will read up on Try trait)

lukaslueg: This is a strictly personal opinion, yet I think both optional function parameters and overloaded constructors are the work of the devil. They seem convenient to the writer, who in fact just puts the burden of an implicit default parameter to the reader.

atroche: I think I understand, but tell me if I'm misunderstanding: is the burden on the reader that if they want to know whether or not defaults are being chosen for them they now need to look up the method signature, rather than having it being clear in the caller's code?

PR #57, opened 2019-02-18
Run cargo test on CI (Linux / Mac only)

atroche:
Seems like tests haven't actually been running 🤦‍♂️


lukaslueg:
I thought they weren't enabled because they didn't yet work on travis :-) There is more work to be done in `.travis.yml`, see [here](https://docs.travis-ci.com/user/gui-and-headless-browsers/#using-the-chrome-addon-in-the-headless-mode)

PR #59, opened 2019-02-18
Improve screenshot API, add example

lukaslueg:
This fixes the test being broken on MacOS for some reason (the screenshot taken from the view is just 1x1 pixel), improves the API (no optional parameters that sometimes get ignored, sometimes not) and adds an executable example.

### DIFF
--- a/src/cdtp/page.rs
+++ b/src/cdtp/page.rs
@@ -13,11 +13,18 @@ pub struct Frame {
     pub unreachable_url: Option<String>,
 }
 
-/// The format a screenshot will be captured in
-#[derive(Debug, Clone, Serialize)]
+#[derive(Debug, Serialize)]
 #[serde(rename_all = "lowercase")]
-pub enum ScreenshotFormat {
+pub(crate) enum InternalScreenshotFormat {
     JPEG,
+    PNG,
+}
+
+/// The format a screenshot will be captured in
+#[derive(Debug, Clone)]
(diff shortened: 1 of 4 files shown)


src/cdtp/page.rs
```diff
@@ -13,14 +13,21 @@ pub struct Frame {
     pub unreachable_url: Option<String>,
 }
 
-/// The format a screenshot will be captured in
-#[derive(Debug, Clone, Serialize)]
+#[derive(Debug, Serialize)]
 #[serde(rename_all = "lowercase")]
-pub enum ScreenshotFormat {
+pub(crate) enum InternalScreenshotFormat {
```
atroche: From my testing this doesn't seem to be necessary, I suppose because `cdtp` is already a public module in `lib.rs`?
```suggestion
pub enum InternalScreenshotFormat {
```

src/cdtp/page.rs
```diff
@@ -73,12 +80,11 @@ pub mod methods {
 
     #[derive(Serialize, Debug)]
     #[serde(rename_all = "camelCase")]
-    pub struct CaptureScreenshot {
-        pub format: Option<super::ScreenshotFormat>,
+    pub(crate) struct CaptureScreenshot {
```
atroche:
```suggestion
    pub struct CaptureScreenshot {
```

PR #60, opened 2019-02-18
Two small fixes

lukaslueg:
The `termcolor`-crate should be unneeded (and is in fact); if downstream wants colored output, they can always do that.

README has been broken by recent changes


atroche:
Thank you!

PR #61, opened 2019-02-18
Add query_wikipedia example

lukaslueg:
A free-standing version of whats in `real_world.rs`.

I noticed that `tab.wait_for_element()` and especially `tab.find_element()` can return an error `NoElementFound` that is no longer public and therefor can't be matched against. We currently can't distinguish if the element is simply not there or if the browser failed in some other way.


lukaslueg:
cc @atroche

PR #62, opened 2019-02-18
Use the builder pattern for launch options

r-darwish:
Using the builder pattern means that the auto detection function will only run if the path was not specified.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ toml = "0.4"
 chrono = "0.4"
 which = "2.0"
 base64 = "0.10"
+derive_builder = "0.7.1"
 
 [target.'cfg(windows)'.dependencies]
 winreg = "0.6"
--- a/README.md
+++ b/README.md
@@ -6,10 +6,10 @@
 [Puppeteer](https://github.com/GoogleChrome/puppeteer) for Rust. It looks a little something like this:
 
 ```rust
-use headless_chrome::{Browser, LaunchOptions};
+use headless_chrome::{Browser, LaunchOptionsBuilder};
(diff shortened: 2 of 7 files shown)


lukaslueg:
Ah, that LaunchOptions-thing was really a pain, thanks!

For what it's worth, my two cents:

* The only time we need a `LaunchOptions`, it's the only parameter to the constructor. We could have `LaunchOptions.browser() -> Result<Browser, Error>` and `LaunchOptions.process() -> Result<Process, Error>` instead of `Browser::new(LaunchOptions)` and `Process::new(LaunchOptions)`. If by the time `browser()` or `process()` is called the executable is not set, it's searched for; if that fails, we return a type-safe error instead. So spinning up a browser could look like `LaunchOptions::default().browser()?.wait_for_initial_tab()?....`.
* I'm not strongly opinionated yet I don't think the extra dependency is worth it for such a tiny struct.

atroche:
@r-darwish thanks, I think this is a real improvement. It also has me wondering how the API would look if we used the Builder pattern for `Browser` itself rather than `LaunchOptions`.

@lukaslueg I think I'd find it more intuitive (as a new user) to create a `Browser` using one of its constructors rather than via a method on the configuration struct, although of course it's highly subjective. I'm open to having both options available.

Bringing in `derive_builder` seems worth it to me.

atroche:
It's a bit of a shame that `.build()` returns `Result<_, String>` though: https://github.com/colin-kiegel/rust-derive-builder/issues/60

Means we can't do this:

```rust
LaunchOptionsBuilder::default().build()?
```

PR #66, opened 2019-02-19
Implement Browser.getVersion

lukaslueg:


src/cdtp/browser.rs
```diff
@@ -0,0 +1,26 @@
+pub mod methods {
+    use crate::cdtp::Method;
+    use serde::{Deserialize, Serialize};
+
+    #[derive(Serialize)]
+    pub struct GetVersion {}
+    #[derive(Deserialize, Clone, Debug)]
+    #[serde(rename_all = "camelCase")]
+    /// Version information returned by `Browser.getVersion`
+    pub struct VersionInformationReturnObject {
```
atroche: Ah, so I've been trying to call all of these as `MethodNameReturnObject`, so that one day if / when we decide to put all this repetition behind macros or a build script it's easier. Like this:

image

But perhaps for this one we could have a public type alias (like `VersionInformation`), if we're going to use it in the public API?

lukaslueg: Yeah, I actually had a public `VersionInformation` with an `Into::into`-implementation to hide the internal type. I removed it, though, because a set of Strings seemed simple enough. There should be no protocol-types leaking through the API in the long run.

A type alias will just compile to the internal type, including in the docs, which kind of defeats this purpose.

atroche: Fair enough, I'm sure it won't be hard to deal with the naming if / when we decide to reduce boilerplate in the `cdtp` mod

lukaslueg: We could have the `cdtp::Method` implementors use a `T` (the publicly visible type) where the actual method call is bound to a `U: Into<T>`. The current code does not change since all `T` implement `Into::into<T>`. Yet the protocol-specific types can implement their own `Into` to convert to the publicly-visible type.

Do you want me to change `VersionInformationReturnObject` being publicly visible?

atroche: Would using that `T` affect the type inference on `Transport.call_method`?

I'm not bothered by `VersionInformationReturnObject` being public. Was just slightly concerned by the break in naming convention.

lukaslueg: Actually yes probably we'd have to change every definition to something like `FooReturnObject as Into<FooReturnObject>` which is less than ideal. I may experiment.

We can always change the type name, there aren't too many people's code getting broken...

Rebase and merge if this looks good to you.

PR #68, opened 2019-02-20
WIP: Implement clip on captureScreenshot

lukaslueg:
Here is an implementation of the `clip` argument on `captureScreenshot`, which allows taking a screenshot of specific regions on the page. The functionality has been hotwired to `tab::Element` so one can do `tab.wait_for_element("div#foobar")?.capture_screenshot(ScreenshotFormat::PNG)?` to get a screenshot of that specific element. See `capture_screenshot_element_box()` for how to do this in a more fine-grained way. Docs, tests and examples were expanded.

Two problems remain:

* `tab::Element` is not public and so it's not documented, neither are `BoxModel`, `ElementQuad` etc. This is the same problem as with `NoElementFound` returned by `tab.find_element()`.
* The integration tests all fail in mysterious and flaky ways. Most of the time they just time out, sometimes `transport` complains about not being able to communicate. The tests seem to be stable if executed `--test-threads 1`. Recent changes broke something?

### DIFF
--- a/examples/take_screenshot.rs
+++ b/examples/take_screenshot.rs
@@ -1,19 +1,28 @@
-//! Create a headless browser, navigate to wikipedia, wait for the page
-//! to render completely, take a screenshot in JPEG-format using 75% quality
 use headless_chrome::{cdtp::page::ScreenshotFormat, Browser, LaunchOptionsBuilder};
 use std::fs;
 
 fn main() -> Result<(), failure::Error> {
+    // Create a headless browser, navigate to wikipedia.org, wait for the page
+    // to render completely, take a screenshot of the entire page
+    // in JPEG-format using 75% quality.
     let options = LaunchOptionsBuilder::default()
         .build()
         .expect("Couldn't find appropriate Chrome binary.");
-    let browser = Browser::new(options).expect("Failed to launch and connect to Chrome");
-    let jpeg_data = browser
-        .wait_for_initial_tab()?
+    let browser = Browser::new(options)?;
+    let tab = browser.wait_for_initial_tab()?;
+    let jpeg_data = tab
         .navigate_to("https://www.wikipedia.org")?
         .wait_until_navigated()?
-        .capture_screenshot(ScreenshotFormat::JPEG(Some(75)), true)?;
+        .capture_screenshot(ScreenshotFormat::JPEG(Some(75)), None, true)?;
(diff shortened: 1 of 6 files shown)


atroche:
Putting `capture_screenshot` on `Element` is neat, and makes for an impressive README example.

I'll have a crack at making the browser, tab and element modules public (while not making so many of their items public).

Thanks for the heads up about the integration tests, I'll have to take another look. Have you found it failing on Travis or on your machine, and on master? Last few Travis builds seem to have passed, except for `cargo fmt` complaint on this PR.

atroche:
Aha, here's one: https://travis-ci.com/atroche/rust-headless-chrome/jobs/179373461

I think the problem is with `[wait_until_navigated](https://github.com/atroche/rust-headless-chrome/blob/3d4378fcb15945f5554a5e29a14288f87f4d6bbd/src/browser/tab/mod.rs#L131)`. I think there's a race condition there, when Chrome finishes navigating before we hit the first `wait_for`. The logic really needs a re-think.

I'll have (another!) read through how Puppeteer manages it, but any help would be much appreciated.

atroche:
P.S. as a work around for now, I'll replace https://github.com/atroche/rust-headless-chrome/blob/d6deb5b67ceb0c702b0087065dbf6a5e827bd612/tests/simple.rs#L80 with a `wait_for_element`. It looks like this is the only test in `simple.rs` relying on `wait_until_navigated`.

atroche:
I think the solution might be to integrate waiting into the `Tab.navigate_to` fn itself, so the CDTP method call to navigate can't out-race "wait for navigation".

lukaslueg:
The other tests failed too so the problem cant be limited to `wait_until_navigated()` ?!

atroche:
From that Travis job I linked to:

image


I'm not sure which other tests you're referring to.

atroche:
Okay, I figured it out. A lot of the tests on this branch drop browser straight away after getting it back from `dumb_server` or `dumb_client`, which kills the chrome process. I suppose we'll need to find a way to keep browser alive while any of its tabs still are, which will probably mean putting browser behind an Arc.

But for now, we can work around it in those tests with this:

https://github.com/atroche/rust-headless-chrome/commit/649538fb6aa453a03152ceecf4c47908101aefde

I'll file an issue.

PR #69, opened 2019-02-20
Add support for preloading extensions

r-darwish:


src/browser/process.rs
```diff
@@ -170,6 +179,14 @@ impl Process {
             args.extend(&["--headless"]);
         }
 
+        let extension_args: Vec<String> = launch_options
+            .extensions
+            .iter()
+            .map(|e| format!("--load-extension={}", e.display()))
```
lukaslueg: There is room for improvement here as Command [arguments](https://doc.rust-lang.org/std/process/struct.Command.html#method.arg) are `AsRef<OsStr>`, not `String`, to begin with. `PathBuf` already is `AsRef<OsStr>`. Going from `PathBuf` via `fmt::Display` to `String` and back to `OsStr` is unnecessary. The current `args`-type also allows to specify arguments which are actually invalid.

r-darwish:
@lukaslueg is this good to merge?

lukaslueg:
My thinking was that `extension_args` should be a `Vec<OsStr>` or similar, there shouldn't be as many type conversions/re-referencing required. Yet this is a minor thing, we can change it later.

Could we also add the test from #42 so we know this PR makes Chrome at least load with this extra argument?

lukaslueg:
Oh and a later PR should at least link to the docs about loading extensions. That is, what das Chrome actually accept here - a zip, a folder ?!

r-darwish:
Done. I changed the test a bit but kept @ravi0213 as the commit author

src/browser/process.rs
```diff
@@ -170,6 +179,14 @@ impl Process {
             args.extend(&["--headless"]);
         }
 
+        let extension_args: Vec<String> = launch_options
+            .extensions
+            .iter()
+            .map(|e| format!("--load-extension={}", e.to_str().unwrap()))
```
lukaslueg: This could be something to the tune of `let a = String::new("--load-extension").into(); a.push(e)` to avoid `e` from going through string.

lukaslueg:
LGTM

atroche:
LGTM

atroche:
(thanks!)

PR #76, opened 2019-02-22
change clippy::stutter->module_name_repetitions

leshow:
Pulled master and got a bunch of clippy errors, after changing this, errors went away. Feel free to ignore if this isn't needed.

edit: Was also 2 more lints left


atroche:
Thanks looks like we need to still allow `stutter` though. I'll add it back :)

atroche:
Maybe I had an old version of `clippy` or something when I ran it on my other machine.

atroche:
looks like travis uses that version too? I'll keep both.

Issue #74, opened 2019-02-21
Browser process shouldn't be dropped while a tab is still in scope


atroche:
https://github.com/atroche/rust-headless-chrome/pull/68#issuecomment-466207288

lukaslueg:
I'm not familiar with how `Browser`, `Process`, `Tab` and `Element` are actually tied together. Yet in principle this should not be a problem if `Tab<'a>` contained a reference to `Browser` (or an `Rc`). So I'm wondering why we have this problem in the first place.

Another example is `wait_for_initial_tab()`, which should return a `&Tab` instead of `Arc<Tab>`

atroche:
Re. your second point, it's because `Browser`s and `Tab`s each have their own event handling loops running (in separate threads). The browser's might receive a “TargetCreated” event (say, when clicking a link opens a popup), and then need to modify its internal list of tabs. While at the same time a user (from the main thread) might wonder how many tabs the browser has open, and want to get a handle to one of them.

Currently Browser and Tab both reference (via an `Arc`) a `Transport`, which is a wrapper around a WebSocket connection and also a `WaitingCallRegistry` that keeps track of in-flight method calls. If we want the browser process to be kept alive, we'll have to make sure that either the entire `Browser` (or at least its child process) is referenced by the tab.

(Side note: I'd like to move away from using mutexes wherever I can and rely on message passing via channels.)

I realise this is hardly an optimal setup. It's overly complicated. But being able to have tabs being controlled in separate threads, and also to have the load of (de)serialization spread across cores is quite important to some work I'm doing. But I'm certain there are better ways to achieve those goals than the current situation!

lukaslueg:
I'm thinking about `Runtime.ScriptId` and `ExecutionContext`. Both are tied to the underlying instance and the underlying instance may go away, rendering the `ScriptId` a useless (possibly dangerous) value; holding on to a `ScriptId` should keep the entire browser session alive. 

My current thinking is to either use a weak pointer (where the event loop has the only strong reference) or - preferably - make `ScripId` something to the tune of `struct<'a> { scriptId, t: PhantomData<&'a Tab> }`. That way `ScriptId` is tied to the `Tab` and can't be used if the `Tab` is dropped.

Any thoughts on this?

atroche:
Do `ScriptId` and `ExecutionContext` truly need to reference all of `Tab`, or just its `Transport`?

Something relevant that's just occurred to me:

[...]

PR #84, opened 2019-02-28
Add BoxModel-helpers

lukaslueg:
Some helper functions on `BoxModel` that are handy when writing tests downstream, to check simple geometry (e.g. "is this element roughtly where it's supposed to be")

### DIFF
--- a/src/browser/tab/element.rs
+++ b/src/browser/tab/element.rs
@@ -49,6 +49,130 @@ impl ElementQuad {
     pub fn aspect_ratio(&self) -> f64 {
         self.width() / self.height()
     }
+
+    /// If the most bottom point of `self` is above the most top point of `other`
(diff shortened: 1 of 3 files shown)


lukaslueg:
Just one bit away from the perfect commit id...

lukaslueg:
CI seems to be unrelated but legit. @atroche Any idea if this is #74 in some form?

atroche:
Looking now

atroche:
@lukaslueg I don't think so, because the test that's hanging is solely dealing with `Process`, and not `Browser` / `Tab`. But I've made a couple of small changes that should help debug if it happens again (and which potentially would've fixed it):
https://github.com/atroche/rust-headless-chrome/pull/85

Still not sure exactly how Process::new() could form an infinite loop.

lukaslueg:
Rebased, merge if you want

Issue #82, opened 2019-02-25
Expose JS line coverage from a tab


atroche:
This is what I'll be adding to the library next, I think. I'll post a proposed API before I carry on implementing it though.

atroche:
What do you think of this API for JS coverage tracking, @lukaslueg @leshow @r-darwish ?

References:
https://pptr.dev/#?product=Puppeteer&version=v1.8.0&show=api-class-coverage
https://chromedevtools.github.io/devtools-protocol/tot/Profiler#method-startPreciseCoverage

As a first pass I'm not too concerned about call counts.

```rust
#[test]
fn returns_actual_coverage() -> Result<(), Error> {
    let server = server_with_html_and_js();
    let browser = Browser::new(LaunchOptionsBuilder::default().build().unwrap()).unwrap();
    let tab: Tab = browser.wait_for_initial_tab()?;

    tab.start_js_coverage();

    let url = format!("http://127.0.0.1:{}", server.port());
    tab.navigate_to(&url)?;

    let coverage_report = tab.take_precise_js_coverage()?;
    let script_coverage = coverage_report.get_by_url(format!("{}/coverage_fixture1.js"))?;

    // see the file coverage_fixture1.js
    const ON_LOAD_LINE_NUMBER: u16 = 2;
    const  ON_CLICK_LINE_NUMBER: u16 = 7;

    let on_load_line_was_covered = script_coverage.get_line_ranges().any(|line_range| {
        line_range.contains(ON_LOAD_LINE_NUMBER);
    });
    assert!(on_load_line_was_covered);

    let on_click_line_was_not_covered = script_coverage.get_line_ranges().all(|line_range| {
        !line_range.contains(ON_CLICK_LINE_NUMBER);
    });
    assert!(on_click_line_was_not_covered);

    let button = tab.wait_for_element("#incrementor")?;
    button.click()?;

    let updated_coverage_report = tab.take_precise_js_coverage()?;
    let updated_script_coverage = coverage_report.get_by_url(format!("{}/coverage_fixture1.js"))?;

    let on_click_line_was_covered = updated_script_coverage.get_line_ranges().any(|line_range| {
        line_range.contains(ON_LOAD_LINE_NUMBER);
    });
    assert!(on_click_line_was_covered);

    let coverage_report = tab.stop_js_coverage()?;

    let coverage_report_result = tab.take_precise_js_coverage();

    // (because we stopped coverage)
    assert!(coverage_report_result.is_err());

[...]

PR #86, opened 2019-03-07
Add wrappers for starting, stopping and taking JS coverage

atroche:
Unlike Puppeteer (and related to this comment: https://github.com/atroche/rust-headless-chrome/issues/82#issuecomment-470167972) I'm not munging the data from Chrome into a friendlier format. We can always add helpers for that later if this causes confusion (or make it the default, to keep more in line with Puppeteer's API as opposed to the Chrome DevTools API).

The test is probably the best entry point for reading this PR, although on the whole I'm doing little more than wrapping raw CDTP methods.

### DIFF
--- a/src/browser/tab/mod.rs
+++ b/src/browser/tab/mod.rs
@@ -1,27 +1,24 @@
-use failure::{Error, Fail};
-use log::*;
-
-use serde;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use std::sync::Mutex;
 
-use super::transport::SessionId;
-use crate::browser::Transport;
-use crate::protocol;
-use crate::protocol::dom;
-use crate::protocol::input;
-use crate::protocol::page;
-use crate::protocol::page::methods::Navigate;
-use crate::protocol::target;
-use crate::protocol::target::TargetId;
-use crate::protocol::target::TargetInfo;
-use crate::protocol::Event;
-
-use super::waiting_helpers::{wait_for, WaitOptions};
+use failure::{Error, Fail};
+use log::*;
+use serde;
 
 use element::Element;
 use point::Point;
+
+use crate::browser::Transport;
+use crate::protocol;
+use crate::protocol::page::methods::Navigate;
+use crate::protocol::target::TargetId;
+use crate::protocol::target::TargetInfo;
+use crate::protocol::Event;
+use crate::protocol::{dom, input, page, profiler, target};
+
+use super::transport::SessionId;
+use super::waiting_helpers::{wait_for, WaitOptions};
 
 pub mod element;
 mod keys;
@@ -379,5 +376,35 @@ impl<'a> Tab {
             script_to_evaluate,
         })?;
         Ok(self)
+    }
+
+    /// Tells Chrome to start tracking which lines of JS have been executed.
+    ///
+    /// Equivalent to hitting the record button in the "coverage" tab in Chrome DevTools.
+    /// See the file `tests/coverage.rs` for an example.
+    pub fn start_js_coverage(&self) -> Result<&Self, Error> {
+        self.call_method(profiler::methods::Enable {})?;
+
+        self.call_method(profiler::methods::StartPreciseCoverage {
+            call_count: None,
+            detailed: Some(true),
+        })?;
+        Ok(self)
+    }
+
+    /// Tells Chrome to stop tracking which lines of JS have been executed.
+    pub fn stop_js_coverage(&self) -> Result<&Self, Error> {
+        self.call_method(profiler::methods::StopPreciseCoverage {})?;
+        Ok(self)
+    }
+
+    /// Returns info about which lines of JS have been executed since you last called this method
+    /// The format of the data is a little unintuitive, see here for details:
+    /// https://chromedevtools.github.io/devtools-protocol/tot/Profiler#type-ScriptCoverage
+    pub fn take_precise_js_coverage(&self) -> Result<Vec<profiler::ScriptCoverage>, Error> {
+        let script_coverages = self
+            .call_method(profiler::methods::TakePreciseCoverage {})?
+            .result;
+        Ok(script_coverages)
     }
 }
--- a/src/browser/tab/mod.rs
+++ b/src/browser/tab/mod.rs
@@ -378,27 +378,54 @@ impl<'a> Tab {
         Ok(self)
     }
 
-    /// Tells Chrome to start tracking which lines of JS have been executed.
+    /// Enables the profiler
+    pub fn enable_profiler(&self) -> Result<&Self, Error> {
+        self.call_method(profiler::methods::Enable {})?;
+
+        Ok(self)
+    }
+
+    /// Disables the profiler
+    pub fn disable_profiler(&self) -> Result<&Self, Error> {
+        self.call_method(profiler::methods::Disable {})?;
+
+        Ok(self)
+    }
+
+    /// Starts tracking which lines of JS have been executed
+    ///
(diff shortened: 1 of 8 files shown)


src/browser/tab/mod.rs
```diff
@@ -380,4 +377,34 @@ impl<'a> Tab {
         })?;
         Ok(self)
     }
+
+    /// Tells Chrome to start tracking which lines of JS have been executed.
+    ///
+    /// Equivalent to hitting the record button in the "coverage" tab in Chrome DevTools.
+    /// See the file `tests/coverage.rs` for an example.
+    pub fn start_js_coverage(&self) -> Result<&Self, Error> {
+        self.call_method(profiler::methods::Enable {})?;
```
lukaslueg: We may want to unhinge the `Enable`-call into a separate public method. It will always result in a clear error message if `Profiler.enable` has not been called, which makes forgetting to call this easy to debug. Unhinging makes it possible to implement `Profile.disable`, which people may want to use e.g. for performance reasons.

src/browser/tab/mod.rs
```diff
@@ -380,4 +377,34 @@ impl<'a> Tab {
         })?;
         Ok(self)
     }
+
+    /// Tells Chrome to start tracking which lines of JS have been executed.
```
lukaslueg: I'd suggest a unified wording of `Start tracking which lines of JS have been executed` instead of `Tells Chrome...`.

src/browser/tab/mod.rs
```diff
@@ -380,4 +377,34 @@ impl<'a> Tab {
...
+        self.call_method(profiler::methods::Enable {})?;
+
+        self.call_method(profiler::methods::StartPreciseCoverage {
+            call_count: None,
+            detailed: Some(true),
+        })?;
+        Ok(self)
+    }
+
+    /// Tells Chrome to stop tracking which lines of JS have been executed.
```
lukaslueg: `Stop tracking which lines of JS have been executed`

src/browser/tab/mod.rs
```diff
@@ -380,4 +377,34 @@ impl<'a> Tab {
...
+    pub fn stop_js_coverage(&self) -> Result<&Self, Error> {
+        self.call_method(profiler::methods::StopPreciseCoverage {})?;
+        Ok(self)
+    }
+
+    /// Returns info about which lines of JS have been executed since you last called this method
```
lukaslueg: The original docs on this seem more detailed: `Collect coverage data for the current isolate, and resets execution counters. Precise code coverage needs to have started.`

src/protocol/profiler.rs
```diff
@@ -0,0 +1,87 @@
+use serde::Deserialize;
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(rename_all = "camelCase")]
+/// Coverage data for a source range.
+pub struct CoverageRange {
+    pub start_offset: u16,
+    pub end_offset: u16,
+    pub count: u16,
```
lukaslueg: The original types for these are `integer`, for which we better use `u64` to prevent overflow, especially on the counter.

src/protocol/profiler.rs
```diff
@@ -0,0 +1,87 @@
+use serde::Deserialize;
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(rename_all = "camelCase")]
+/// Coverage data for a source range.
+pub struct CoverageRange {
+    pub start_offset: u16,
```
lukaslueg: The original docs for these items are

`JavaScript script source offset for the range start.`
`JavaScript script source offset for the range end.`
`Collected execution count of the source range.`

Which are always helpful

src/browser/tab/mod.rs
```diff
@@ -380,4 +377,34 @@ impl<'a> Tab {
...
+    /// Tells Chrome to start tracking which lines of JS have been executed.
+    ///
+    /// Equivalent to hitting the record button in the "coverage" tab in Chrome DevTools.
+    /// See the file `tests/coverage.rs` for an example.
+    pub fn start_js_coverage(&self) -> Result<&Self, Error> {
+        self.call_method(profiler::methods::Enable {})?;
+
+        self.call_method(profiler::methods::StartPreciseCoverage {
+            call_count: None,
+            detailed: Some(true),
```
lukaslueg: It should be documented that `isBlockCoverage` will always be true.

src/protocol/profiler.rs
```diff
@@ -0,0 +1,87 @@
...
+
+/// JS line coverage information for a single script
+/// See https://chromedevtools.github.io/devtools-protocol/tot/Profiler#type-ScriptCoverage
+#[derive(Debug, Clone, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ScriptCoverage {
+    pub script_id: String,
+    /// Either the name or URL of a script loaded by the page
+    pub url: String,
+    /// Functions contained in the script that has coverage data
```
lukaslueg: `have`

tests/coverage.rs
```diff
@@ -0,0 +1,122 @@
...
+        .find(|function_coverage| function_coverage.function_name == "button.onclick")
+        .unwrap();
+
+    let fn_range = on_click_function_coverage.ranges.first().unwrap();
+    assert_eq!(fn_range.count, 0);
+
+    let button = tab.wait_for_element("#incrementor")?;
+    button.click()?;
+
+    let updated_script_coverages = tab.take_precise_js_coverage()?;
```
lukaslueg: There is an implicit counter reset here?

lukaslueg:
I clicked on changes-required but this lgtm overall. The only change I'd feel more strongly about is to separate the `enable`-call and provide the corresponding `disable`

src/protocol/profiler.rs
atroche: So what they call integer is actually an i32, and if you go over that you get an error. I figured for non negative numbers a u16 makes just as much sense, but perhaps it'd be better to just use i32s everywhere when dealing with the protocol, just in case.

lukaslueg: I've used u64 elsewhere before. Maybe a type alias `JsInt = i32` and `JsUInt = u32` so those members read `pub count: JsUInt` and make clear why this particular type

lukaslueg: The thing with "`i32` everywhere" is that one is semi-implicitly in signed-arithmetic-land then, which does not make sense e.g. for a counter.

atroche: You only get the error of you go over the max value of i32

lukaslueg: I was thinking more about going to -1 instead of getting an underflow abort

atroche: Ah sorry, that comment was in response to your comment about having used u64. Agreed re type aliases, except that using u32 for unsigned JS int would allow for sending numbers too large

lukaslueg: That is also true but without resorting to special integer types one thing has to give :)

atroche: True! I'll make that change.

src/browser/tab/mod.rs
atroche: Yep, will do.

src/protocol/profiler.rs
atroche: This is copied from the official docs, and I'm guessing the 'has' is referring to the script rather than the functions?

tests/coverage.rs
atroche: I tried to cover that in the comment below this line about only getting "updates", but I'll add a line to the docs of this function.

PR #88, opened 2019-03-09
Refactor waiting-helper

lukaslueg:
This moves the waiting helper into the public. My use-case is a test that needs to wait until a Node has disappeared from the DOM, the opposite of `wait_for_element`.


src/browser/tab/mod.rs
```diff
@@ -174,17 +161,17 @@ impl<'a> Tab {
...
     pub fn wait_for_element(&'a self, selector: &'a str) -> Result<Element<'a>, Error> {
-        self.wait_for_element_with_custom_timeout(selector, 15_000)
+        self.wait_for_element_with_custom_timeout(selector, std::time::Duration::from_secs(15))
     }
 
     pub fn wait_for_element_with_custom_timeout(
         &'a self,
         selector: &'a str,
-        timeout_ms: u64,
+        timeout: std::time::Duration,
```
lukaslueg: Note that this is technically a breaking change in the public API

src/util.rs
```diff
@@ -0,0 +1,61 @@
...
+        }
+    }
+
+    /// Wait until the given predicate returns `Some(G)` or timeout arrives.
+    pub fn until<F, G>(&self, predicate: F) -> Result<G, Timeout>
+    where
+        F: FnMut() -> Option<G>,
+    {
+        let mut predicate = predicate;
+        let start = Instant::now();
```
lukaslueg: This also fixes a bug in the current helper, which uses `SystemTime`, which is not monotonic...

lukaslueg:
I have some tests which *almost* succeed on Travis. The test that is the motive for this PR fails, however.  [The test](https://github.com/lukaslueg/macro_railroad_ext/blob/e153cce5ee125d82147429e469c548cf03a730de/extension/integration_test/src/main.rs#L125) causes an SVG to be recomputed and a certain Node should disappear from the DOM. This succeeds locally [but fails on travis.](https://api.travis-ci.org/v3/job/504043757/log.txt)

Is there something fishy going on with `find_element()` ? Is the answer to `find_element()` cached somehow and the test get's stalled data?

atroche:
Reading now

atroche:
Looks good to me! Feel free to merge.

I can't think of a reason why that test you linked to would fail on Travis but not locally. Does it fail consistently? We're definitely not doing any caching on our end. Are the tests run with the same Chrome / Chromium revision?

atroche:
Gonna merge now so I can branch off it more easily.

PR #90, opened 2019-03-09
Add find / wait for elements methods

atroche:
Still unsure about the whole `wait` vs `find` part of the API, and also the necessity of the "custom timeout" version, and also the `found_via_selector` field, but those can wait for another time.

### DIFF
--- a/src/browser/tab/mod.rs
+++ b/src/browser/tab/mod.rs
@@ -18,6 +18,7 @@ use crate::protocol::{dom, input, page, profiler, target};
 use crate::{protocol, util};
 
 use super::transport::SessionId;
+use crate::protocol::dom::Node;
 
 pub mod element;
 mod keys;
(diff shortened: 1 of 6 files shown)


src/browser/tab/mod.rs
```diff
@@ -185,17 +187,31 @@ impl<'a> Tab {
...
     }
 
+    pub fn wait_for_elements(&'a self, selector: &'a str) -> Result<Vec<Element<'a>>, Error> {
+        debug!("Waiting for element with selector: {}", selector);
+        util::Wait::with_timeout(Duration::from_secs(15))
+            .until(|| {
+                if let Ok(elements) = self.find_elements(selector) {
+                    if elements
+                        .iter()
+                        .all(|element| element.get_midpoint().is_ok())
```
lukaslueg: Why is this here?

lukaslueg:
The `found_via_selector`-item seems superfluous and, afaics, adds an unnecessary lifetime restriction to `Element`. It probably should go some time.

lukaslueg:
The whole `find`/`wait` thing boils down to the fact that this is a synchronous API and waiting is the only thing one can do. There is no way to fundamentally solve concurrent events in a synchronous API. 

I'd suggest to simply remove all `find` variants and only use `wait`.

src/browser/tab/mod.rs
atroche: It really shouldn't be! Removed from here and from `wait_for_element`.

PR #94, opened 2019-03-13
Lift lifetime restrictions on Tab

lukaslueg:
The lifetime is not used globally in the entire `impl` and the restriction can therefore be lifted.

### DIFF
--- a/src/browser/tab/mod.rs
+++ b/src/browser/tab/mod.rs
@@ -48,7 +48,7 @@ pub struct NavigationFailed {
     error_text: String,
 }
 
-impl<'a> Tab {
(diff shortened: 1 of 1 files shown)


lukaslueg:
CI failed again because the screenshot tests are too strict (color is slightly off). I'll update that.

On a second thought: Shouldn't we simply remove the `found_via_selector` thing on `Element` and let go of the lifetime? It seems like a useless thing to have, despite the extra-logging.

atroche:
Yep, I'm not attached to it at all.

lukaslueg:
Closed in favor of #101

PR #101, opened 2019-03-14
Remove selector from Element

lukaslueg:


src/browser/tab/mod.rs
```diff
@@ -242,19 +240,13 @@ impl<'a> Tab {
...
-        for node_id in &node_ids {
-            elements.push(Element::new(&self, *node_id, selector)?)
+            return Err(NoElementFound {}.into());
         }
 
-        Ok(elements)
+        node_ids
+            .into_iter()
+            .map(|node_id| Element::new(&self, node_id))
+            .collect()
```
lukaslueg: Note that one can `.collect()` into a `Result<T<U>, E>` where `T: FromIterator<Item=Result<U, E>>`, a highly useful shortcut.

atroche:
Cheers

PR #102, opened 2019-03-15
Add Browser.getWindowBounds, setWindowBounds

lukaslueg:


lukaslueg:
It's somewhat unclear where this belongs: The protocol part is in the Browser domain, yet we always need a `TargetID` so this is always bound to a tab. I hope the docs make clear that changes to the bounds affect all tabs...

lukaslueg:
Any idea why this fails on MacOS?

atroche:
No idea re. failure on MacOS, sorry.

lukaslueg:
This needs more investigation, the actual behavior seems highly platform dependent and not at all what the docs suggest.

atroche:
@lukaslueg you could include it only for Linux. Seems better than having it for no platforms.

atroche:
It works fine on my Mac. I think the problem is just with Travis, so I've disabled the test on Mac and merged this.

PR #107, opened 2019-03-19
introduced remote chrome service support and printToPDF functionality

Dowwie:


Dowwie:
Need some clarification on why travis is failing..

src/protocol/page.rs
```diff
@@ -92,6 +127,14 @@ pub mod events {
 pub mod methods {
     use crate::protocol::Method;
     use serde::{Deserialize, Serialize};
+    use super::PrintToPdfOptions;
+
+
+    #[derive(Debug, Deserialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct ReturnDataObject {
```
atroche: For consistency, could you rename this `PrintToPdfReturnObject`? When it comes time to use a build script or macros to generate these protocol structs, having a consistent naming pattern will help the transition.

examples/print_to_pdf.rs
```diff
@@ -0,0 +1,36 @@
...
+fn main() -> Result<(), failure::Error> {
+    // Create a headless browser, navigate to index.html file, wait for the page
+    // to render completely, render a PDF of the page and save it to the filesystem.
+
+    // webSocketDebuggerUrl is obtained by GET http request to 
+    // http://{chromeservicehost}{port}/json/version
+    // and pass an empty-value host param: {host: ""} in header 
+    let file_path = format!("file://{}", 
+                            env::args().nth(1).expect("Must provide path/to/file/index.html"));
+    let debug_ws_url = "ws://127.0.0.1:9222/devtools/browser/14804b82-0392-43be-b20f-d75678460e43";
```
lukaslueg: Is this `UUID` generic?

Dowwie: @lukaslueg that uuid is unique to this example and has to be replaced and recompiled in order for the example to work.  I've updated my PR with this part externalized, requiring it as a parameter

tests/simple.rs
```diff
@@ -152,6 +152,16 @@ fn capture_screenshot_jpeg() -> Result<(), failure::Error> {
     Ok(())
 }
 
+#[test]
+fn test_print_file_to_pdf() -> Result<(), failure::Error> {
+    logging::enable_logging();
+    let (_, browser, tab) = dumb_server(include_str!("./pdfassets/index.html"));
+    let local_pdf = tab.wait_until_navigated()?
+                       .print_to_pdf(None)?;
+    assert_eq!(true, local_pdf.len() > 1000);  // an arbitrary size
```
lukaslueg: This should read `assert!(!local_pdf.is_empty())`. PDFs start with `b"%PDF"`, which would be even better and self-documenting.

Dowwie: @lukaslueg a pdf of an empty page can render to > 0 sized.  1000 ensures this isn't an empty pdf

lukaslueg: I'm not sure if I understand your point. If the test ensures that the buffer is not empty, it should read `assert!(!local_pdf.is_empty())` and avoid arbitrary values. An even better test would be to ensure that the PDF magic bytes are there, something to the tune of `assert!(local_pdf.starts_with(b"PDF%"))`

Dowwie: The pdf magic bytes is a good idea.  I've added it.  The non-zero size test you are recommending does not consider empty pdf generation.  To simulate what I am referring to:

1. cargo build --example print_to_pdf
2. copy the print_to_pdf from the target directory to /tests/pdfassets
3. start a google-chrome-stable headless instance, copy the ws url provided
4. change the working dir to /tests/pdfassts and call " ./print_to_pdf {ws_url} index.html"

This creates a rust.pdf that is 894 bytes but has no content -- it is empty.  local_pdf is 894 bytes yet has no content.

atroche:
Thanks, great work!

PR #110, opened 2019-03-20
Add changelog, bump version to 0.1.4

atroche:
Let me know what you think!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "headless_chrome"
-version = "0.2.0"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "headless_chrome"
-version = "0.1.3"
+version = "0.2.0"
```
lukaslueg: I think if this gets bumped to `0.2`, we should have docs everywhere. I suggest we bump to `0.1.4` or something

PR #113, opened 2019-03-20
Add fetcher as a default feature

leshow:
I killed the progressbar and removed `indicatif`

`reqwest`, `zip`, and `directories` are optional, but default. so @lukaslueg you should be fine to run w/o bringing in any tokio deps by using:

```
[dependencies.headless_chrome]
default-features = false
``` 

Come to think of it, I need to add that to the readme

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,11 +24,10 @@ toml = "0.4"
 chrono = "0.4"
 base64 = "0.10"
 derive_builder = "0.7.1"
-reqwest = "0.9"
-directories = "1.0"
-indicatif = "0.11"
-zip = "0.5"
 which = "2.0"
+reqwest = { version = "0.9", optional = true }
+directories = { version = "1.0", optional = true }
+zip = { version = "0.5", optional = true }
 
 [target.'cfg(windows)'.dependencies]
 winreg = "0.6"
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -43,4 +42,5 @@ name = "headless_chrome"
 path = "src/lib.rs"
 
 [features]
+default = [ "reqwest", "directories", "zip" ]
```
lukaslueg: Could we make this a named feature in itself instead of dispersed in `default`? Something like `fetch`

leshow: yeah of course, good idea.

leshow:
Seems like the CI is erroring on trying to run some non-rust code from the readme file. Anyone know how to prevent that?

README.md
```diff
@@ -59,7 +59,7 @@ If you get errors related to timeouts, you likely need to enable sandboxing eith
 
 By default, `headless_chrome` will download a compatible version of chrome to `XDG_DATA_HOME` (or equivalent on Windows/Mac). This behaviour can be optionally turned off, and you can use the system version of chrome (assuming you have chrome installed) by disabling the default feature in your `Cargo.toml`:
 
-```
```
lukaslueg: The `toml` is required because CI will pull in the README as an external doc and make sure that all code examples actually compile

atroche:
LGTM!

Issue #106, opened 2019-03-19
Allow users to register callbacks for specific browser / tab events


Dowwie:
wait_until_X logic ought to be used for edge cases rather than the standard for this library.  Instead, callbacks should be registered and called when certain events are triggered by chrome.  For instance, when loadEventFired is sent by chrome, call print_to_pdf(..).

atroche:
I agree that it would be good to allow adding callbacks for various events. It hasn't come up in my work yet, so I haven't put any thought into designing an API for it.

That being said, while I'm open to have an alternative fully-asynchronous API, I prefer using the synchronous one, because otherwise you end up with browser tests with a fair bit of boilerplate in them, like all the `and_then`s in [this example from fantoccini](https://github.com/jonhoo/fantoccini#examples).

Is it okay if I change the title of this issue to "Allow users to add callbacks for specific browser / tab events"? Or, if it's more what you're after, "implement asynchronous API"?

Interestingly, `fantoccini` has an inverse issue: https://github.com/jonhoo/fantoccini/issues/43

Dowwie:
@atroche I agree with your recommended title change, although a fully async api would also be great.

jianglibo:
callbacks? I cannot find term callback in Tokio, but Futures. I tried all and_thens, skip_whiles all of them end with messes. When I stop use these future combinators and implement own futures, things begin to look easier. Think of Browser as a source of raw messages(all message) of a stream. then Page as a more specific message source, finally your own page which consumes Page stream, and sends method to.  You should start writing code as Tokio way, starting off with a future. the code below demonstrates this idea:
```rust
impl Future for MyPage {
    type Item = ();
    type Error = failure::Error;

    fn poll(&mut self) -> Poll<(), Self::Error> {
        loop {
            if let Some(value) = try_ready!(self.chrome_page.poll()) {
                    match &mut self.state {
                        MyPageState::Start => {
                            info!("*** Start ***");
                            if let PageMessage::DocumentAvailable = value {
                                self.state = MyPageState::WaitingNode;
                                self.chrome_page.find_node(self.node_id);
                            }
                        }
                        MyPageState::WaitingNode => {
                            info!("*** WaitingNode ***");
                            if let PageMessage::FindNode(maybe_selector, nd) = value {
                                if Some(self.node_id.to_string()) == maybe_selector {
                                    info!("got node {:?}", nd);
                                    self.state = MyPageState::WaitElement;
                                }
                            }
                        }
[...]

PR #115, opened 2019-03-21
changed default sleep from 100 to 5 ms

Dowwie:


lukaslueg:
If we go low on the default sleep time, I'd be great if we had some form of bounded exponential backoff, e.g. 5ms, 25ms, 100ms, 250ms, 250ms, 250ms....

Dowwie:
It's not clear to me whether exponential backoff fits this scenario.  If this were backing off from hammering a Chrome service, then maybe, but this sleeps between checking whether state changed.

atroche:
@Dowwie some of the uses of `Wait` do relatively expensive operations, like multiple RPC calls. Why not just change the timeout used for the navigation check? (That was your specific use case, IIRC from chat)

The backoff idea seems pretty sensible to me.

Dowwie:
@atroche I see.  Thanks for clarifying that io is involved.  I think exponential can take off a little too quickly so can we start it at 5ms?  Changing just for nav is up to you.

atroche:
Is there a specific problem you ran into that you're trying to solve? Let me know if I can help.

Dowwie:
@atroche adopting backoff may involve changing sleep's type from Duration to ExponentialBackoff, depending on what you prefer for the library.  Changing sleep to Duration would introduce breaking changes that you have to be comfortable with.  Can you think of an alternative?

Backoff in a nutshell:

backoff library:   https://docs.rs/backoff/0.1.3/backoff/

```rust
use backoff::backoff::Backoff;
use backoff::ExponentialBackoff;

let mut backoff = ExponentialBackoff::default();
backoff.max_elapsed_time = None;

let time = backoff.next_backoff();
```

Dowwie:
@atroche next step requires your feedback about changing types

atroche:
@Dowwie the Wait struct could contain an enum of either Duration or ExponentialBackoff, and use the latter by default but the former by users specifying a timeout.

Dowwie:
I'm closing this PR.  Changing sleep or incorporating backoff are tactical solutions.  A headless-chrome puppeteer is intended to be asynchronous.  An async library isn't sleeping nor backing off but rather responding when tasks are ready.

Dowwie:
see https://github.com/atroche/rust-headless-chrome/issues/106

PR #116, opened 2019-03-21
Remove examples/real_world and dependencies

lukaslueg:
The example has fallen behind in terms of code style and does not provide much value any more. It also pulls in toml for the sole purpose of reading a secrets file most people wont bother to have.


atroche:
Good idea

PR #118, opened 2019-03-22
Fix windows version of archive_name and default_executable

### LINKED ISSUE #117: Not compiling on windows
Hi, I get the following errors when a crate I'm using trys to compile this one:

```shell
error[E0425]: cannot find value `revision` in this scope
   --> C:\Users\dkolsoi\.cargo\registry\src\github.com-1ecc6299db9ec823\headless_chrome-0.1.4\src\browser\fetcher.rs:250:12
    |
250 |         if revision.as_ref().parse::<u32>()? > 591_479 {
    |            ^^^^^^^^ help: a local variable with a similar name exists: `_revision`

error[E0425]: cannot find function `get_chrome_path_from_registry` in this scope
   --> C:\Users\dkolsoi\.cargo\registry\src\github.com-1ecc6299db9ec823\headless_chrome-0.1.4\src\browser\mod.rs:367:29
    |
367 |         if let Some(path) = get_chrome_path_from_registry() {
    |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in this scope
help: possible candidate is found in another module, you can import it into scope
    |
1   | use crate::browser::process::get_chrome_path_from_registry;
    |
```

It seems as though `_revision` is the actual variable name in the former and `get_chrome_path_from_registry` isn't imported in the latter

TheDan64:
Fixes #117

### DIFF
--- a/src/browser/fetcher.rs
+++ b/src/browser/fetcher.rs
@@ -247,7 +247,7 @@ fn archive_name<R: AsRef<str>>(_revision: R) -> Result<&'static str, Error> {
     #[cfg(windows)]
     {
         // Windows archive name changed at r591479.
-        if revision.as_ref().parse::<u32>()? > 591_479 {
+        if _revision.as_ref().parse::<u32>()? > 591_479 {
             Ok("chrome-win")
         } else {
             Ok("chrome-win32")
--- a/src/browser/fetcher.rs
+++ b/src/browser/fetcher.rs
@@ -233,21 +233,25 @@ where
     }
 }
 
-fn archive_name<R: AsRef<str>>(_revision: R) -> Result<&'static str, Error> {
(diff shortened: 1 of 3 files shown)


src/browser/fetcher.rs
```diff
@@ -247,7 +247,7 @@ fn archive_name<R: AsRef<str>>(_revision: R) -> Result<&'static str, Error> {
     #[cfg(windows)]
     {
         // Windows archive name changed at r591479.
-        if revision.as_ref().parse::<u32>()? > 591_479 {
+        if _revision.as_ref().parse::<u32>()? > 591_479 {
```
lukaslueg: Underscored variables should not actually be used. I Suggest to rename the input variable to `revision` and do `drop(revision); Ok(..)` in the linux and macos paths.

atroche:
Cheers :)

Issue #93, opened 2019-03-11
Feature-gate the fetcher


lukaslueg:
I was very surprised by the number of dependencies that the `fetcher`-module pulls in, including tokio, parking_lot, hyper and openssl. Especially `indicatif`,  which is far fetched for a library of this kind. This greatly increases compilation times.

As a matter of personal taste, I consider the whole progress bar thing gimmicky at best. It does, however, potentially garble `stdout/stderr` of *any* program that uses `headless_chrome` as a transient dependency.

Could we please feature-gate the whole fetcher module?

atroche:
I think that's a good idea. Any thoughts @leshow ?

leshow:
Sounds good to me. It would be a pretty easy change.

Also, if you really don't like the progress bar, we could always just remove that and just use log.

leshow:
Could be a good opportunity to split off protocol as a separate module using some kind of workspace configuration also. Any thoughts on that? 

Do we want fetcher enabled or disabled by default?

lukaslueg:
My two cents on this would be to 1) remove the progress-bar-thing entirely and 2) make the downloading-behaviour explicit instead of implicit. That is, with the feature disabled everything just works as long as Chrome is there. If the feature is enabled, one can use the fetcher module to download a working revision. `LaunchOptions` has to work the same way with/without the feature.

leshow:
I think taking a que from puppeteer may be helpful here? By default puppeteer downloads a bundled chrome, IMO we shouldn't have a default behavior that can fail, like running system chrome that happens to be incompatible, or in an env without chrome.

Puppeteer gives you environment variables to override that behavior, we could have something like that, or put using the system chrome behind a feature like you mentioned.

lukaslueg:
I've no strong feelings about whether or not its the default as long as the dependency chain of ~80 extra dependencies can be disabled :)

leshow:
I can take a look at doing this sometime this week, if you like. Any other comments @atroche ?

atroche:
I'm also okay with it being the default and being able to disable it. I think as long as it's reflected in the docs it's fine. Thanks!

tazjin:
> IMO we shouldn't have a default behavior that can fail, like running system chrome that happens to be incompatible

Downloading a binary from the internet is also a default behaviour that can fail (if e.g. you're running a system with dynamically linked binaries in non-Debian/RedHat-conformant locations).

[...]

PR #119, opened 2019-03-23
Replace reqwest with ureq

leshow:
#93 let's close this

This came up a few times; reqwest brings in a huge transitive dependency. It was suggested to replace with `ureq`. It looks cross platform? I don't have any environments other than linux to test.

If someone could try on mac & windows that'd be nice.

Also, prints DL size in MiB now instead of bytes

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ tempfile = "3"
 base64 = "0.10"
 derive_builder = "0.7.1"
 which = "2.0"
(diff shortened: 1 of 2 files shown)


leshow:
The other thing that bothers me is the time that unzipping takes, there's an issue for that for `zip-rs` https://github.com/mvdnes/zip-rs/issues/88

atroche:
Awesome, thank you.

PR #120, opened 2019-03-24
Added await_promise argument to call_js_fn method

### LINKED ISSUE #112: Add ability to await promises in call_js_fn method
There is an optional parameter for `Runtime.callFunctionOn` in devtools protocol
```
awaitPromise boolean
```
I'd like to extend `CallFunctionOn` struct with `await_promise` field in protocol methods, and add one more argument to `call_js_fn`.

ibaryshnikov:
Closes #112 

Added one more field to `CallFunctionOn` struct, and argument to `call_js_fn` method
Added tests


atroche:
:+1:

PR #123, opened 2019-04-23
Add window size option

iwatakeshi:
This adds the `--window-size` option to the argument list. [Two](https://travis-ci.com/iwatakeshi/rust-headless-chrome/builds/109332816) of the Travis builds [are failing](https://travis-ci.com/iwatakeshi/rust-headless-chrome/jobs/195028831#L733) but it seems irrelevant to the changes I made. If this is something I need to take care of, please let me know.

### DIFF
--- a/src/browser/mod.rs
+++ b/src/browser/mod.rs
@@ -123,7 +123,11 @@ impl Browser {
     }
 
     pub fn get_process_id(&self) -> Option<u32> {
-        self.process.as_ref().map(|process| process.get_id())
+        if let Some(process) = &self.process {
+            Some(process.get_id())
+        } else {
+            None
+        }
     }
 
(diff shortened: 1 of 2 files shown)


lukaslueg:
Thanks for the PR!

Personally, I'd prefer to not change default behaviour if this option is not set. That said, the default behaviour should be to not pass any arg, instead of passing 1920x1080.

The Travis failings are due to clippy getting updated. I'd be helpful if you could fix them in this PR, even though unrelated. You can basically do what clippy suggests to do.

iwatakeshi:
👍 understandable.

iwatakeshi:
@lukaslueg Would this be good? Thanks!

atroche:
Awesome, thanks @iwatakeshi!

PR #124, opened 2019-04-26
Add sandbox option

iwatakeshi:
This adds the `--no-sandbox` option. While it's not wise to use this option, there are times where it's needed such as using chrome in a docker container. By default, sandbox is enabled so the behavior is the same as before.


atroche:
Cheers mate!

PR #126, opened 2019-05-05
Add chrome paths

iwatakeshi:
This adds additional paths to `default_executable()`. It also adds a check for an environment variable that may contain a path to the executable. 

I wasn't sure if `$CHROME` would be a good variable. I thought `$BROWSER` was too generic and may imply that Firefox or another browser can be accepted. If you like me to change it to `$BROWSER`, please let me know. Thanks.

### DIFF
--- a/src/browser/mod.rs
+++ b/src/browser/mod.rs
@@ -348,13 +348,11 @@ impl Drop for Browser {
 /// Returns the path to Chrome's executable.
 /// 
 /// If the `CHROME` environment variable is set, `default_executable` will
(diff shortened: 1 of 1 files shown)


lukaslueg:
@atroche are you seeing this? I think this is good to go if squashed.

@iwatakeshi Could you add some docs to `default_executable()` which strategies the function will try in which order? Something to the tune of "If the CHROME-env variable is defined, it is used. Otherwise the following filenames are searched for......."

PR #128, opened 2019-06-13
Add implicit scroll into view for elements

sdimkov:
### Description
This PR adds an implicit scroll into view for elements. It is necessary to ensure a smooth flow of execution. The scroll is triggered before any action and only if the element is partially or fully out of view.

The JS code is based on Puppeteer's [implementation](https://github.com/GoogleChrome/puppeteer/blob/6860d73212d2a36fd789a9244667be574f396164/lib/JSHandle.js#L162-L186).

### Tested on
- [x] Mac
- [x] Linux
- [ ] Windows

### DIFF
--- a/src/browser/tab/element.rs
+++ b/src/browser/tab/element.rs
@@ -1,4 +1,4 @@
-use failure::Error;
+use failure::{Error, Fail};
 use log::*;
 
 use super::point::Point;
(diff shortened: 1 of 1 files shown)


atroche:
Cheers mate. Just running tests locally now (because there seems to be a problem with CI unrelated to this PR).

I think this is a good idea, seeing as it's what Puppeteer does by default.

atroche:
Fixed a test to do with Chrome's code coverage tracking (it wasn't accounting for the possibility of anonymous injected scripts in its expectations).

Thanks again.

PR #129, opened 2019-06-16
Fix stalled wait-for-element calls

sdimkov:
There is an odd issue with `wait_for_element` and `wait_for_elements` methods:

If an unexpected error occurs while you wait for some element (for example the browser gets closed / killed) you would expect that the blocking call to `wait_for_element` will fail immediately and provide a meaningful error (for example `RemoteError { code: -32602, message: "No session with given id" }`).

Instead what happens is that you keep on waiting for your call to finish and when it finally does - you get the standard `Timeout` error so you can't respond to the actual problem at hand.

This PR addresses the issue above, by implementing a stricter until method that can explicitly allow certain errors while failing and reporting the rest.

### DIFF
--- a/src/browser/tab/element.rs
+++ b/src/browser/tab/element.rs
@@ -2,6 +2,7 @@ use failure::{Error, Fail};
 use log::*;
 
 use super::point::Point;
+use crate::browser::tab::NoElementFound;
 use crate::protocol::dom;
 use crate::protocol::page;
 use crate::protocol::runtime;
@@ -228,10 +229,13 @@ impl<'a> Element<'a> {
     /// We use these two when making various calls to the API because of that.
     pub fn new(parent: &'a super::Tab, node_id: dom::NodeId) -> Result<Self, Error> {
         if node_id == 0 {
-            return Err(super::NoElementFound {}.into());
+            return Err(NoElementFound {}.into());
         }
 
-        let backend_node_id = parent.describe_node(node_id)?.backend_node_id;
+        let backend_node_id = parent
+            .describe_node(node_id)
+            .map_err(NoElementFound::map)?
+            .backend_node_id;
 
         let remote_object_id = {
             let object = parent
--- a/src/browser/tab/mod.rs
+++ b/src/browser/tab/mod.rs
@@ -64,20 +64,22 @@ pub struct NavigationFailed {
     error_text: String,
 }
 
-fn handle_find_errors(error: Error) -> Error {
-    match error.downcast::<RemoteError>() {
-        Ok(remote_error) => {
-            match remote_error.message.as_ref() {
-                // This error is expected and occurs while the page is still loading,
-                // hence we shadow it and respond the element is not found
-                "Could not find node with given id" => NoElementFound {}.into(),
+impl NoElementFound {
+    pub fn map(error: Error) -> Error {
+        match error.downcast::<RemoteError>() {
+            Ok(remote_error) => {
+                match remote_error.message.as_ref() {
+                    // This error is expected and occurs while the page is still loading,
+                    // hence we shadow it and respond the element is not found
+                    "Could not find node with given id" => Self {}.into(),
 
-                // Any other error is unexpected and should be reported
-                _ => remote_error.into(),
+                    // Any other error is unexpected and should be reported
+                    _ => remote_error.into(),
+                }
             }
+            // Return original error if downcasting to RemoteError fails
+            Err(original_error) => original_error,
         }
-        // Return original error if downcasting to RemoteError fails
-        Err(original_error) => original_error,
     }
 }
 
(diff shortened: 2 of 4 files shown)


sdimkov:
p.s. A regression test that confirms the above issue is resolved may be reasonable to add as well.

But I'm not familiar enough with the testing codebase to figure out how and where to implement it.

atroche:
Thanks @sdimkov, I'm trying to add a test for this now.

atroche:
Okay, I created a test for this in `simple.rs`:
```rust
#[test]
fn wait_for_element_returns_unexpected_errors_early() -> Result<(), failure::Error> {
    logging::enable_logging();
    let (server, browser, tab) = dumb_server(include_str!("simple.html"));
    let downcast_error_result = tab
        // empty selector will get a 'RemoteError' from Chrome because the selector is invalid
        .wait_for_element("")
        .unwrap_err()
        .downcast::<RemoteError>();

    assert!(downcast_error_result.is_ok());
    Ok(())
}
```

But also noticed that these other tests are failing (despite passing on master):
```
    call_js_fn_async_resolved
    call_js_fn_async_unresolved
    reload
    test_print_file_to_pdf
```

I don't have time to debug them right now, is there a chance you could take a look?

sdimkov:
Sure, I’ll give them a try today.

Are the tests currently green and maintained on master? Last time I checked master was failing locally for me (osx).

atroche:
Ah, they reliably pass on my Linux machine, and while I don't have access to a Mac at the moment for a long time they passed on OS X too. Have been meaning to investigate the problem with CI for some time. Maybe it's just an issue with the OS X build...

sdimkov:
Well, if the CI was green, it would be the single source of truth and therefore maintaining the tests on all other platforms would be of lesser importance.

atroche:
Ah yeah, at the moment it runs on both Linux and OS X.

sdimkov:
@atroche I added a potential fix here: https://github.com/atroche/rust-headless-chrome/pull/130

atroche:
I'm finding that on some runs the tests I mentioned above fail, so I'm guessing there's a timing issue. Have some time now to investigate a little more.

atroche:
So, if the page hasn't fully loaded and someone calls `wait_for_element`, the previous behaviour was to keep trying anyway (despite getting back `RemoteError { code: -32000, message: "Could not find node with given id" }`), and the `QuerySelector` method call would eventually return the right thing when the page loaded.

Seeing as `strict_until` allows you to ignore specific errors, I should just make sure that the above `RemoteError` becomes a `NoElementFound`, hence preserving the previous allow-the-page-to-keep-loading behaviour (which IIRC is what Puppeteer does).

atroche:
Ah weird, it seems that the `Could not find node with given id` message is actually referring to the `node_id` of the `document` which we get via `Tab.get_document`. We might still want to make `strict_until` ignore that message, but perhaps we can find a better way.

sdimkov:
The fix should be in `find_element` I think.

sdimkov:
Just `.map_err` to `NoElementFound` here:

atroche:
This modification to `find_element` gets the tests passing reliably:

```
        let node_id = {
            let root_node_id = self.get_document()?.node_id;
            let result = self.call_method(dom::methods::QuerySelector {
                node_id: root_node_id,
                selector,
            });

            match result {
                Ok(node) => node.node_id,
                Err(e) => {
                    let remote_error_result = e.downcast::<RemoteError>();
                    match remote_error_result {
                        Ok(remote_error) => {
                            // this error message seems to only occur when you query the DOM
                            // before the page has fully loaded. also we can't use `code` to
                            // distinguish it, because different kinds of errors (e.g. invalid
                            // selector) have the same code.
                            if remote_error.message == "Could not find node with given id" {
                                return Err(NoElementFound {}.into());
                            } else {
                                return Err(remote_error.into());
                            }
                        }
                        Err(other_error) => {
                            return Err(other_error.into());
                        }
                    }
                }
            }
        };
```

The reason we need to check the error message is that for some errors (and even some `RemoteErrors` (aka ones returned from Chrome) we don't want to return `NoElementFound` and keep retrying, we want to immediately return the error (e.g. in the case of the invalid selector).

I'm sure there's a nicer way of expressing this than my nested matches, btw, so feel free to make the code nicer.

sdimkov:
Is the code `-32000` distinct for this specific error? If so we should match based on it instead of by message.

atroche:
It's not. For example, using an invalid selector returns:

```
RemoteError { code: -32000, message: "DOM Error while querying" }
```

Bit of a shame, really!

sdimkov:
@atroche another look? CI is green.

sdimkov:
Thanks @atroche !

P.s. Also a random note from me - I see you use `Rabase and Merge` option in GitHub. I'd advise to use the standard `Merge` (with a merge commit) instead.

Having the merge commits for each feature/topic branch results in much better history down the road. And as a bonus, not pulling the commits into master separately lets you keep the original commit author, GPG key and other history metadata.

atroche:
Sure thing, will change to using `Merge`.

PR #133, opened 2019-06-27
Add response handling callback for tabs.

### LINKED ISSUE #132: Inspect response contents
Puppeteer allows to inspect the content of responses pretty easily. 

This is currently not possible with this crate, as far as I can tell?

(Specifically, I need to read the JSON body of api responses)

theduke:
Adds a `Tab::enable_response_handler` method that installs a callback that will be called for each finished network request.

The callback receives the response info, and also a additional method that retrieves the payload. (I'm not so sure about that one, but I think it's the most straight-forward solution)

A test is included. 

Closes #132

### DIFF
--- a/src/browser/tab/mod.rs
+++ b/src/browser/tab/mod.rs
@@ -43,6 +43,18 @@ pub type RequestInterceptor = Box<
         + Sync,
 >;
 
+#[rustfmt::skip]
+pub type ResponseHandler = Box<
+    Fn(
+            protocol::network::events::ResponseReceivedEventParams,
+            &dyn Fn() -> Result<
+                protocol::network::methods::GetResponseBodyReturnObject,
+                failure::Error,
+            >,
+        ) + Send
+        + Sync,
+>;
+
 /// A handle to a single page. Exposes methods for simulating user actions (clicking,
 /// typing), and also for getting information about the DOM and other parts of the page.
 pub struct Tab {
@@ -52,6 +64,7 @@ pub struct Tab {
     navigating: Arc<AtomicBool>,
     target_info: Arc<Mutex<TargetInfo>>,
     request_interceptor: Arc<Mutex<RequestInterceptor>>,
+    response_handler: Arc<Mutex<Option<ResponseHandler>>>,
 }
 
 #[derive(Debug, Fail)]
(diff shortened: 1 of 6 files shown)


atroche:
Thank you! It's extremely impressive to me that you knocked that out so quickly.

theduke:
Thanks. Luckily the code-base is pretty straight-forward and easy to navigate.
;)

atroche:
Oh, would you mind getting the build green before I merge?

theduke:
Formatting errors... I ran `cargo fmt` and just pushed again.

theduke:
Oh man, I actually found a bug in `rustfmt` that formats code in a way that causes a `internal error: left behind trailing whitespace` error.

theduke:
Not sure what's going on with the tests, they are passing locally for me.

atroche:
Hmm, I get warnings when I run `cargo clippy --all --tests --examples` on your branch on my Ubuntu box. Not sure why those would be errors on Travis.

src/browser/tab/mod.rs
```diff
@@ -43,6 +43,18 @@ pub type RequestInterceptor = Box<
         + Sync,
 >;
 
+#[rustfmt::skip]
+pub type ResponseHandler = Box<
```
lukaslueg: Shouldn't this be a trait?

theduke: Yeah, it should.

I was just sticking to how the intercept handler is defined.

Happy to introduce a trait instead.

The bigger question for me is: the tests are a bit flaky in general, and especially my new test. Sometimes passes, sometimes fails. Something fails in the wait for navigation handling. 

I won't have time for a few days to dig into this so any advice is appreciated.

atroche: Running tests on master over and over now, hopefully I can trigger a failure.

atroche: Here's my little notebook I'm keeping (mainly for myself) as I work through this: https://paper.dropbox.com/doc/Navigation-in-headless_chrome--AgCNa0JHNG4ouDv5R_m~2Ip6AQ-viXQEuMOi3HXHeD58oZYi

atroche: @lukaslue do you think this should block a merge and release? and do you think `Tab.request_interceptor` should be a trait too? I'm not sure of the best way to do it, otherwise I would now.

atroche: https://github.com/atroche/rust-headless-chrome/issues/137

PR #134, opened 2019-06-29
Bump dependencies

lukaslueg:
This bumps some dependencies to recent versions, tests seem to work fine.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,13 +18,13 @@ serde_derive = "1"
 failure = "0.1"
 log = "0.4"
 env_logger = "0.6"
-rand = "0.6"
+rand = "0.7"
 tempfile = "3"
 base64 = "0.10"
 derive_builder = "0.7.1"
 which = "2.0"
-ureq = { version = "0.9", optional = true }
-directories = { version = "1.0", optional = true }
+ureq = { version = "0.11", optional = true }
+directories = { version = "2.0", optional = true }
 zip = { version = "0.5", optional = true }
 
 [target.'cfg(windows)'.dependencies]
@@ -35,7 +35,6 @@ chrono = "0.4"
 tiny_http = "0.6"
 png = "0.14"
 jpeg-decoder = { version = "0.1", default_features = false }
-base64 = "0.10.1"
 
 [lib]
 name = "headless_chrome"
--- a/tests/simple.rs
+++ b/tests/simple.rs
@@ -440,7 +440,7 @@ fn get_script_source() -> Result<(), failure::Error> {
 
     tab.wait_until_navigated()?;
 
-    sleep(Duration::from_millis(100));
+    sleep(Duration::from_millis(500));
 
     let script_coverages = tab.take_precise_js_coverage()?;
 


atroche:
Probably worth bumping our own version too. What do you reckon?

lukaslueg:
Should probably wait after #133 gone in?!

atroche:
Cool

sdimkov:
> Probably worth bumping our own version too.

Sorry to bother, could I give a random advice here? I'd recommend following [Semver](https://semver.org/).

Following it doesn't only mean having `X.Y.Z` version number but also following the semantics for each of those numbers. I say this because I noticed last release was made as a patch (`0.1.4` -> `0.1.5`) while it contained a lot of new features. Patch release should contain _only_ backwards compatible bug fixes, while last release had a lot of new features.

So a new version containing #133 and #134 should most likely be a _minor_ release, not a _patch_ (`0.1.5` -> `0.2.0`)

I hope I don't sound like a dick. Best regards!

atroche:
@sdimkov thanks! I appreciate the advice, I think you're right.

lukaslueg:
FWIW I feel more relaxed about this: We are way below `1.0` and [anything goes](https://doc.rust-lang.org/cargo/reference/manifest.html#the-version-field). Adding features does not break builds that do not rely on these features, especially if `struct` definitions do not change. While integers do come cheap, I'd rather not pretend to be semver compatible while in fact we have no way to *guarantee* that anyway;

lukaslueg:
@atroche  merge if you want

atroche:
@lukaslueg okay, will think over the versioning another time :)

sdimkov:
> we have no way to guarantee that anyway

I assume you mean, that we can't guarantee backward compatibility. I'm curious what do you see as necessary for this project to guarantee it? (especially given at some point this project should reach `1.0.0` milestone and then semver will be kinda mandatory)

Also even if we simply agree that there is no guarantee, how is it better to keep releasing patch versions compared to minor versions?

p.s. Maybe this discussion is better suited for chat but I don't see people to be active there.

lukaslueg:
AFAICs the crate-ecosystem doesn't really have a trustworthy solution to this, as there are a lot of caveats. For example, things like `pub fn<'...>(...) -> impl Foobar` can carry "hidden" lifetime constraints in the return type, which could break semver compatibility without touching the signature.

At the end of the day, once a crate reaches `1.0` - separating from the crowd of `<1.0 aka "yeah I'm stable, kind of, except when I'm not` - one has to be very carefuly with every single PR.

atroche:
@lukaslueg what are the harms of bumping minor versions for new features? Personally, my only gripe is aesthetic (in that it would look and feel strange to have (e.g.) `v0.37.0`). A concrete advantage to adopting semvar is that changes to the version number will communicate their contents better to a wider audience (whereas at the moment one really needs to read the changelog to get a feel for how much has changed).

lukaslueg:
It's mainly an argument for convenience. Yet, keep in mind that people usually bind their dependency on minor versions (`^0.37`); bumping the minor version when adding non-semver-breaking changes tends to leave downstream behind.

I'd recommend to start writing a more formal `CHANGES.md`, so changes can be better tracked between released versions.

atroche:
I've been following this format: https://keepachangelog.com/en/1.0.0/. What kind of formality are you thinking?

PR #136, opened 2019-07-06
Implement default for browser

atroche:
I'm not sure if this is a good idea, or if I should just rely on a test helper to reduce the duplication involved in creating a headless browser with `default_executable` in so many tests.

Let me know if you think this isn't a sensible / convenient default.


lukaslueg:
Personally, I don't think impls of `Default` should ever panic and we have two `unwraps()` in here. People are going to use `Default` as they should, see that it works locally but don't see that it causes crashes in library code for others because there is no error path. The stdlib [strongly expects](https://github.com/rust-lang/rust/issues/61129) `Default` not to panic.

atroche:
@lukasleug cool, I won't merge this. Do you think it's worth having a method that does the same thing though, for convenience?

lukaslueg:
Yes, something to the tune of

```rust
impl Browser {
    fn default() -> Result<Self, Error> {
        let launch_options = LaunchOptionsBuilder::default()
            .path(Some(default_executable()?))
            .build()?;
        Self::new(launch_options)
    }
}
```

PR #138, opened 2019-07-08
Disable unused features

lukaslueg:
This disables some unused features, cutting down compile times ever so slightly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ regex = "1"
 serde = { version = "1", features = ["serde_derive"] }
 serde_json = "1"
 serde_derive = "1"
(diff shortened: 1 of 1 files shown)


lukaslueg:
Where did the CI go ? :-)

atroche:
there it is. don't have time right now to look at what that failure is unfortunately.

lukaslueg:
Could you restart CI, the error seems to be transient.

atroche:
Hmm, I did, but no luck

lukaslueg:
Turns out there is a bug in `png` so we can't disable the feature anyway. CI is due to legit clippy warning after 1.36, can be fixed separately

PR #145, opened 2019-07-30
Added idle_browser_timeout to LaunchOptions with a default of 30 seconds

jeprojects:


jeprojects:
@atroche I just noticed this error now after 30 seconds.

`ERROR [headless_chrome::browser::transport] Transport loop got a timeout while listening for messages (Chrome #Some(1506))`

I assume that the same timeout should be applied to both `browser` and `transport`?

jeprojects:
Tested this in a project of mine and all is fine.

Running the tests on mac without these changes come up with 5 failures so that might be why tests are failing with Travis CI?

atroche:
@jeprojects ah, looks like it's actually `rustfmt`. can you run merge the latest master into your branch and then run this?

```
    cargo fmt --all -- --check;
    cargo clippy --all --tests --examples;
```

By the way, I tried running the test suite on Mac and only had problems with that one "response handler" test. Could you report the others as an issue if they're still there?

atroche:
Oh, and also if / when you post the issue, could you mention what Chromium version it is? Could be that.

atroche:
Just ran it against Chromium 78.0.3870.0 (latest as of yesterday) with no problems.

jeprojects:
Looks like clippy doesn't like this function with too many arguments:

```
warning: this function has too many arguments (8/7)

fn handle_incoming_messages(
        messages_rx: Receiver<protocol::Message>,
        waiting_call_registry: Arc<WaitingCallRegistry>,
        listeners: Listeners,
        open: Arc<AtomicBool>,
        conn: Arc<WebSocketConnection>,
        shutdown_rx: Receiver<()>,
        process_id: Option<u32>,
        idle_browser_timeout: Duration,
    )
```

I have run rustfmt and done another commit.

atroche:
Could you use [this](https://stackoverflow.com/a/55402970)?

atroche:
p.s. we don't have to do it now, but I'm curious as to how you'd refactor that. Make a separate struct responsible for message handling?

jeprojects:
Yeah perhaps a separate struct. I'm open to suggestions.

jeprojects:
Tests which still fail for me on mac using Chrome Version 75.0.3770.142 

```
thread 'capture_screenshot_element' panicked at 'assertion failed: sum_of_errors(&buf[i * 4..(i + 1) * 4], &[0x33, 0x22, 0x11, 0xff])

thread 'capture_screenshot_element_box' panicked at 'assertion failed: dbg!(sum_of_errors ( & buf [ 0 .. 4 ] , & [ 0x22 , 0x11 , 0x33 , 0xff ] ))

thread 'capture_screenshot_jpeg' panicked at 'assertion failed: sum_of_errors(&buf[0..4], &[0x11, 0x22, 0x33])

thread 'capture_screenshot_png' panicked at 'assertion failed: sum_of_errors(&buf[0..4], &[0x11, 0x22, 0x33, 0xff])
```

atroche:
huh, interesting. how much is `sum_of_errors` off by I wonder? might be able to fix just by upping the constant it's being compared against.

@lukaslueg you've dealt with these before, right?

jeprojects:
For `capture_screenshot_element_box` - `sum_of_errors` = 17

atroche:
Well, if you can be bothered, you could set those constants to higher values. Luckily it doesn't happen on CI.

Feel free to merge when green!

jeprojects:
Looks like that fixed it.

PR #147, opened 2019-07-31
extract devtools-proto crate

flier:
Extract the chrome DevTools protocol to a standalone crate like `devtools-proto`, and we can test and reuse the core DevTools protocol in other tools.


atroche:
Nice one! Could I ask that it be called something like `chrome_devtools_protocol`?

Also, what does the dev workflow look like for when you're developing the two crates simultaneously? Is there an easy way to use the working version of the protocol crate without publishing it?

atroche:
@lukaslueg would be interested to get your thoughts on this

flier:
@atroche I think you may try to use [the patch section](https://doc.rust-lang.org/cargo/reference/manifest.html#the-patch-section)

lukaslueg:
I don't have a strong opinion on this but my first impression is that `headless_chrome` could just re-export the binding-part.

flier:
1. `headless_chrome` has a dozen dependencies, which is no need for a standalone tools use `chrome_devtools_protocol `.
2. `headless_chrome` only use a part of [devtools protocol](https://chromedevtools.github.io/devtools-protocol/1-3), it's no sense to add an unused API to `headless_chrome`.
3. `headless_chrome` use a handmade API for the [devtools protocol](https://chromedevtools.github.io/devtools-protocol/1-3), it should be generated base on [the latest PDL file](https://github.com/ChromeDevTools/devtools-protocol/tree/master/pdl). I'm working on a [pdl](https://github.com/flier/rust-pdl) crate to parse it, and we could automatic generate whole API in a `chrome_devtools_protocol` crate.

atroche:
I think splitting protocol out is a good idea, I just want to have a go with developing alongside these changes before I merge. Hope that doesn't inconvenience you!

flier:
Sure, I created [a temporary repo](https://github.com/flier/rust-chrome-devtools-protocol) to maintain the automatic generated binding types, and will merge back to `chrome_devtools_protocol ` when you ready.

flier:
Some generated Rust [binding files](https://gist.github.com/flier/5a3eab35793b5fee35279adc313040a3) with [chrome-devtools-protocol](https://github.com/flier/rust-chrome-devtools-protocol).

flier:
Now, we have two working example, including [print_to_pdf.rs](https://github.com/flier/rust-chrome-devtools-protocol/blob/master/examples/print_to_pdf.rs) and [take_screenshot.rs](https://github.com/flier/rust-chrome-devtools-protocol/blob/master/examples/take_screenshot.rs), which implements `CallSite/EventSource` and `AsyncCallSite/AsyncEventSource` to dispatch the method call as websocket message.

@atroche would you have time to review the design, if it's ok, I will start to migrate `rust-headless-chrome` protocol and transport.

```rust
    let target_id = browser.wait_for_initial_tab().unwrap().target_id;
    let session_id = dbg!(browser.attach_to_target(target_id, None)?);

    let mut session = Session {
        endpoint: browser,
        session_id,
    };

    session.enable()?;
    session.set_lifecycle_events_enabled(true)?;

    let (frame_id, loader_id, err_msg) =
        dbg!(session.navigate(opt.url.to_string(), None, None, None))?;

    if let Some(msg) = err_msg {
        bail!("navigate failed, {}", msg);
    }

    dbg!(session.wait_until_navigated(frame_id, loader_id)).unwrap();

    let (data, _stream) = session.print_to_pdf(Default::default())?;

    fs::write("output.pdf", data)?;
```
It still use `futures` v0.1, do we need migrate to use `async/await`?
```rust
    let client = ClientBuilder::new(ws_uri.as_str())?
        .async_connect_insecure()
        .map_err(move |err| err.context(format!("connect {}", ws_uri)).into())
        .map(move |(client, _headers)| Endpoint::new(client))
        .and_then(|endpoint| endpoint.set_discover_targets(true))
        .and_then(|(_, endpoint)| endpoint.wait_for_initial_tab())
        .and_then(|(target_info, endpoint)| {
            endpoint.attach_to_target(target_info.unwrap().target_id, None)
        })
        .map(|(session_id, endpoint)| Session {
            endpoint,
            session_id,
        })
        .and_then(|session| session.enable())
        .and_then(|(_, session)| session.set_lifecycle_events_enabled(true))
        .and_then(|(_, session)| session.navigate(target_url, None, None, None))
        .and_then(|((frame_id, loader_id, err_msg), session)| {
            if let Some(msg) = err_msg {
                Either::A(future::err(format_err!("fail to navigate, {}", msg)))
            } else {
                Either::B(session.wait_until_navigated(frame_id, loader_id))
            }
        })
        .and_then(|(_, session)| {
            session.capture_screenshot(
                Some(page::CaptureScreenshotRequestFormat::Png),
                None,
                None,
                None,
            )
        })
        .and_then(|(data, _)| {
            fs::write("output.png", &data).map_err(|err| err.context("write image").into())
        })
        .map(|_| ())
        .map_err(|err| eprintln!("error: {:#?}", err));
```

atroche:
Hey @flier, I'm not keen on rewriting the crate to be async, but I'm happy to use the generated types if they fit with the current implementation.

See some of my thoughts on the change to async:

https://www.reddit.com/r/rust/comments/aqgsxk/announcing_the_headless_chrome_crate_puppeteer/eggd0e9/
https://github.com/atroche/rust-headless-chrome/issues/106#issuecomment-474605243

It's not clear to me that users of the crate would see benefits of an async implementation until someone needs to make many, many concurrent connections, which is think is (currently) a pretty obscure use case.

flier:
@atroche I agree that most of user could use `sync` mode for most scenes at page level. the `async` mode will be more suitable for advance user who need complicated multi-user scenes, fuzzing test, percolation test or performance test.

Anyway, I think we could start from the generated types first, and may revisit the `async` mode API after the `async/await` stable.

atroche:
@flier cool. FWIW my own use case involves driving 15 browsers at once (doing fuzzing of web apps), and I'm not running into any problems with the sync implementation. Maybe I will soon, when I try to scale it up more.

d4h0:
> Could I ask that it be called something like chrome_devtools_protocol?

"Chrome" is a trademark of Google, [as far as I know](https://www.google.com/permissions/trademark/trademark-list/). It's probably not a good idea to use trademarks in package names.

Besides that, the DevTools protocol is implemented in browsers besides Chrome, so `devtools_protocol` probably would be a better name.

mdrokz:
Closing this since i have implemented https://github.com/mdrokz/auto_generate_cdp into rust-headless-chrome to generate the CDP protocol at build time.

fasterthanlime:
> Closing this since i have implemented https://github.com/mdrokz/auto_generate_cdp into rust-headless-chrome to generate the CDP protocol at build time.

I'm a tiny bit late to this party, but this breaks building dependents of headless_chrome in nix:  https://hachyderm.io/@fasterthanlime/111168253420369615 (which makes me very sad, I don't want to maintain a fork of headless_chrome)

mdrokz:
> > Closing this since i have implemented https://github.com/mdrokz/auto_generate_cdp into rust-headless-chrome to generate the CDP protocol at build time.
> 
> I'm a tiny bit late to this party, but this breaks building dependents of headless_chrome in nix: https://hachyderm.io/@fasterthanlime/111168253420369615 (which makes me very sad, I don't want to maintain a fork of headless_chrome)

Hey i can add a feature flag in both headless_chrome and auto_generate_cdp to take local json files present in the project this is how i build the docs (doc builds are sandboxed as well) i should have done this from the start.

fasterthanlime:
> Hey i can add a feature flag in both headless_chrome and auto_generate_cdp to take local json files present in the project this is how i build the docs (doc builds are sandboxed as well) i should have done this from the start.

That would be extremely helpful! I forgot about docs.rs’s own sandbox.

mdrokz:
> > Hey i can add a feature flag in both headless_chrome and auto_generate_cdp to take local json files present in the project this is how i build the docs (doc builds are sandboxed as well) i should have done this from the start.
> 
> That would be extremely helpful! I forgot about docs.rs’s own sandbox.

Hey i have a pushed a new version of auto_generate_cdp with the offline feature flag i will integrate it into headless_chrome tomorrow, can you test it locally on your side with a test project ?

fasterthanlime:
> Hey i have a pushed a new version of auto_generate_cdp with the offline feature flag i will integrate it into headless_chrome tomorrow, can you test it locally on your side with a test project ?

My test project is proprietary, but using [a quick fork of rust-headless-chrome](https://github.com/fasterthanlime/rust-headless-chrome/tree/offline) that use auto_generate_cdp's offline feature, I was able to confirm that it all builds with nix (through [crane](https://crane.dev/)), which is what I wanted to do in the first place!

mdrokz:
> > Hey i have a pushed a new version of auto_generate_cdp with the offline feature flag i will integrate it into headless_chrome tomorrow, can you test it locally on your side with a test project ?
> 
> My test project is proprietary, but using [a quick fork of rust-headless-chrome](https://github.com/fasterthanlime/rust-headless-chrome/tree/offline) that use auto_generate_cdp's offline feature, I was able to confirm that it all builds with nix (through [crane](https://crane.dev/)), which is what I wanted to do in the first place!

That's great I will integrate it in headless_chrome today and push a new release glad to see it's working! Thanks for testing it.

PR #150, opened 2019-08-03
Implement Runtime.evaluate

jeprojects:
Basic implementation of Runtime.evaluate based of `call_js_fn`

### DIFF
--- a/src/browser/tab/mod.rs
+++ b/src/browser/tab/mod.rs
@@ -665,8 +665,6 @@ impl<'a> Tab {
     }
 
     /// Evaluates expression on global object.
-    ///
-    /// See Devtools protocol doc: https://chromedevtools.github.io/devtools-protocol/tot/Runtime#method-evaluate
     pub fn evaluate(
         &self,
         expression: &str,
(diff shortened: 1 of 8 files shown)


src/browser/tab/mod.rs
```diff
@@ -663,6 +663,28 @@ impl<'a> Tab {
             .call_method(protocol::debugger::methods::GetScriptSource { script_id })?
             .script_source)
     }
+
+    /// Executes javascript
+    ///
+    /// See Devtools protocol doc: https://chromedevtools.github.io/devtools-protocol/tot/Runtime#method-evaluate
```
lukaslueg: The docs should not just refer to an external syntax.

src/browser/tab/mod.rs
```diff
@@ -663,6 +663,28 @@ impl<'a> Tab {
             .call_method(protocol::debugger::methods::GetScriptSource { script_id })?
             .script_source)
     }
+
+    /// Executes javascript
```
lukaslueg: The first sentence from the original docs is "Evaluates expression on global object.", which is a much better description thatn "Executes javascript".

jeprojects: Thanks.. Made the changes

src/browser/tab/mod.rs
jeprojects: Fixed

atroche:
Cheers, good work

PR #155, opened 2019-08-15
feat(Log): add basic support for Log domain, update Runtime types

RyuuGan:
And log Domain basic support

Also file log is called `logs` which I don't like because module `log` is already used.

### DIFF
--- a/src/protocol/logs.rs
+++ b/src/protocol/logs.rs
@@ -7,27 +7,27 @@ pub mod events {
     #[serde(rename_all = "lowercase")]
     pub enum  LogEntrySource {
         XML,
-        JAVASCRIPT,
-        NETWORK,
-        STORAGE,
-        APPCACHE,
-        RENDERING,
-        SECURITY,
-        DEPRECATION,
-        WORKER,
-        VIOLATION,
-        INTERVENTION,
-        RECOMMENDATION,
-        OTHER,
+        JavaScript,
+        Network,
+        Storage,
+        AppCache,
+        Rendering,
+        Security,
+        Deprecation,
+        Worker,
+        Violation,
+        Intervention,
+        Recommendation,
+        Other,
     }
 
     #[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
     #[serde(rename_all = "lowercase")]
     pub enum  LogEntryLevel {
-        VERBOSE,
-        INFO,
-        WARNING,
-        ERROR,
+        Verbose,
+        Info,
+        Warning,
+        Error,
     }
 
     /// Log entry for a logs
--- a/src/protocol/logs.rs
+++ b/src/protocol/logs.rs
@@ -1,11 +1,11 @@
 pub mod events {
+    use crate::protocol::runtime::methods::{RemoteObject, StackTrace};
+    use crate::protocol::types::JsInt;
     use serde::{Deserialize, Serialize};
-    use crate::protocol::runtime::methods::{StackTrace, RemoteObject};
-    use crate::protocol::types::{JsInt};
 
     #[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
     #[serde(rename_all = "lowercase")]
-    pub enum  LogEntrySource {
+    pub enum LogEntrySource {
         XML,
         JavaScript,
         Network,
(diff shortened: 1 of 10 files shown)


src/protocol/logs.rs
```diff
@@ -0,0 +1,212 @@
...
+    #[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
+    #[serde(rename_all = "camelCase")]
+    pub struct LogEntry {
+        pub source: LogEntrySource,
+        pub level: LogEntryLevel,
+        pub text: String,
+        /// Timestamp when this entry was added
+        /// Number of milliseconds since epoch
+        /// See https://chromedevtools.github.io/devtools-protocol/tot/Runtime#type-Timestamp
+        pub timestamp: f64,
```
lukaslueg: Maybe we want something from `chrono` here so there is no ambiguity.

src/protocol/runtime.rs
```diff
@@ -25,19 +26,90 @@ pub mod methods {
...
+
+    /// Call frames for assertions or error messages
+    /// See https://chromedevtools.github.io/devtools-protocol/tot/Runtime#type-StackTrace
+    #[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+    #[serde(rename_all = "camelCase")]
+    pub struct StackTrace {
+        description: Option<String>,
+        call_frames: Vec<CallFrame>,
+        // parent: Option<StackTrace>,
+        /// Experimental
```
lukaslueg: Could you expand on the comment, giving some words about what is experimental and why.

RyuuGan: Yea, I'll write expand the comment about `parentId` and add a link to Chrome DevTools docs to this

https://chromedevtools.github.io/devtools-protocol/tot/Runtime#type-StackTraceId

src/protocol/logs.rs
RyuuGan: Basically I used the same type for timestamp as in `ResponseReceivedEventParams` from `network.rs`, but in chrome it means a little bit different, but represents still with the same `number` type.

atroche: I'm open to using chrono for these fields later on, but at the moment the other timestamps are all floats, so I don't think this should block the PR

src/protocol/logs.rs
```diff
@@ -0,0 +1,212 @@
+pub mod events {
+    use serde::{Deserialize, Serialize};
+    use crate::protocol::runtime::methods::{StackTrace, RemoteObject};
+    use crate::protocol::types::{JsInt};
+
+    #[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
+    #[serde(rename_all = "lowercase")]
+    pub enum  LogEntrySource {
+        XML,
```
atroche: I'd prefer if these were in `UpperCamelCase` to match the style of enum variants in the rest of the project and also [these guidelines](https://rust-lang-nursery.github.io/api-guidelines/naming.html).

src/protocol/logs.rs
```diff
@@ -0,0 +1,212 @@
...
+        entry: LogEntry,
+    }
+
+    #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+    #[serde(rename_all = "camelCase")]
+    pub struct EntryAddedEvent {
+        pub params: EntryAddedParams,
+    }
+
+    #[test]
```
atroche: Thanks for adding tests! Really appreciate it.

src/protocol/runtime.rs
```diff
@@ -25,19 +26,94 @@ pub mod methods {
...
+        DATAVIEW,
+    }
+
+    #[derive(Deserialize, Serialize, Debug, Clone, PartialEq)]
     #[serde(rename_all = "camelCase")]
     pub struct RemoteObject {
         #[serde(rename = "type")]
-        pub object_type: String,
-        pub subtype: Option<String>,
+        pub object_type: RemoteObjectType,
```
atroche: Nice!

src/protocol/logs.rs
atroche: heh, even though I happened to comment on the `XML` line which is one of the few that wouldn't need to change ;)

RyuuGan: Sure, I'll refactor the Enums, thanks for the link to guidelines.

atroche:
Okay, still having CI problems on Mac (unrelated to your work). Feel free to merge when you've got it passing on Linux :)

src/protocol/logs.rs
RyuuGan: Refactored to `UpperCamelCase`, also added forgotten `pub` in one of the enums. Also ran formatting with `cargo fmt`

PR #161, opened 2019-08-20
Add a default_timeout for Tab

terry90:
Quality of life improvement PR.
Instead of writing everywhere:
```rust
wait_for_element_with_custom_timeout(selector, std::time::Duration::from_secs(5));
wait_for_element_with_custom_timeout(selector1, std::time::Duration::from_secs(5));
...
```
(which is a bit verbose)

We can now write:
```rust
tab.set_default_timeout(std::time::Duration::from_secs(5))
wait_for_element(selector);
wait_for_element(selector1);
```

It also allows `wait_for_elements` timeout to be changed, prior to this, only `wait_for_element` timeout was configurable

### DIFF
--- a/src/browser/tab/mod.rs
+++ b/src/browser/tab/mod.rs
@@ -1,6 +1,5 @@
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::Arc;
-use std::sync::Mutex;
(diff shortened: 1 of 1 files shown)


atroche:
Nice one, thanks

PR #162, opened 2019-08-20
Improve Runtime Domain, Add EventListeners callbacks.

RyuuGan:
Adds `Runtime.enable` and `Runtime.disable` methods, so you can receive `runtime` events.

Adds event listener to listen to the events (see `tests/events_listeners.rs` for more details). 

Relates #106

### DIFF
--- a/src/browser/tab/mod.rs
+++ b/src/browser/tab/mod.rs
@@ -14,7 +14,7 @@ use crate::browser::Transport;
 use crate::protocol::dom::Node;
 use crate::protocol::page::methods::Navigate;
 use crate::protocol::target::{TargetId, TargetInfo};
-use crate::protocol::{dom, input, logs, network, page, profiler, target, Event, RemoteError};
+use crate::protocol::{dom, input, logs, network, page, profiler, runtime, target, Event, RemoteError};
 use crate::{protocol, protocol::logs::methods::ViolationSetting, util};
 
 use super::transport::SessionId;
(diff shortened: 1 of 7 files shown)


atroche:
What a coincidence, was thinking about how to do event listeners yesterday.

I'd like there to be a way to remove listeners after adding them, and also to listen to a specific event instead of to all of them.

Would you be interested in implementing that? I'm happy to make the changes too.

Also feel free to is push back if you disagree.

RyuuGan:
Hi @atroche,

I was thinking about exectly the same things. To remove event listeners is quite easy you just remove it from vector by Refernce, but I didn't find a good way to listen to a specific event. I don't know how to use current enum to subscribe to specific event, so I was thinking to add different enum with all the same events names, but I didn't find this good. 

If it will be good to create the separate enum, it is posible to create a method `from_event` that will return the the EnumType. And then instead of vector we can use the `HashMap` where key will be the event name from the new event enum, and the values are the vector with event listeners.

What do you think about it?

atroche:
Hmm, yeah. I think part of the problem is that `Event` is an enum rather than a trait. If we could use trait objects, we'd be able to add specific listeners. The reason `Event` isn't a trait at the moment is because serde doesn't work well with trait objects. However, I've recently stumbled upon [typetag](https://github.com/dtolnay/typetag), which may be the solution we're looking for.

I'm going to try it out.

atroche:
Note that what it builds under the hood is similar to our `protocol::Event` enum:

image

atroche:
Actually, I'm not sure trait objects is the way to go. If we have:

```rust
type Listener = Box<dyn Fn(dyn Event) -> Fallible<()>>;
```

Then that listener won't be able to access the specific params of its event, only the ones common to all events.

We might have to have separate functions for each event: `tab.on_request_intercepted` etc.

@lukaslueg @r-darwish any ideas?

atroche:
Or we could just let people add callbacks for all events and match on them themselves, as @RyuuGan proposed originally.

atroche:
Just had confirmation from two sources (emailing a friend and asking on Discord) that it's not really possible to do `tab.add_listener`, so @RyuuGan if you'd like to implement the `remove_listener` method I'd be happy to merge this :)

RyuuGan:
@atroche, I was a little bit busy to add remove functionality, but finally I did it.

atroche:
@RyuuGan ah, you legend. I'll take a look on Monday or Tuesday I reckon!

PR #173, opened 2019-09-25
Tabs additional functions

RyuuGan:
Add next functionality to tabs:

1. `Tab#close_target`  - closes the tab.
2. `Tab#close_with_unload` - tries to close the tab, but emits `beforeunload` event.
3. `Tab#close(boolean)` - helper function that chooses one of the functions above depending on the boolean parameter.
4. `Tab#activate` - activates current tab (makes it focus).

### DIFF
--- a/src/browser/mod.rs
+++ b/src/browser/mod.rs
@@ -329,6 +329,14 @@ impl Browser {
                             }
                             Event::TargetDestroyed(ev) => {
                                 trace!("Target destroyed: {:?}", ev.params.target_id);
+                                let mut locked_tabs = tabs.lock().unwrap();
+                                let pos = locked_tabs
+                                    .iter()
+                                    .position(|tab| *tab.get_target_id() == ev.params.target_id);
+
+                                if let Some(idx) = pos {
+                                    locked_tabs.remove(idx);
+                                }
                             }
                             _ => {
                                 let mut raw_event = format!("{:?}", event);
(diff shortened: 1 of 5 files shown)


atroche:
Thanks! I'm going to try clearing Travis's caches to see if that helps the build not error.

atroche:
@RyuuGan huh, seems like there are some legit failures. Do they happen on your machine?

RyuuGan:
@atroche I checked in Windows 10 and Mac OS x 10.14. All the tests pass.

What is the chrome version of the container? Maybe this is the problem?

RyuuGan:
Looks like I have to wait more (and subrscribe for events to wait until the event from chrome fired (TargetDestroyed), I'll check it out soon. Because container is slower then machine

RyuuGan:
The only error left is:

```
failures:
    src/browser/tab/mod.rs - browser::tab::Tab::get_title (line 939)
test result: FAILED. 9 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
error: test failed, to rerun pass '--doc'
```

But I didn't touched this docs, any ideas?

atroche:
Ah well, let's not worry about that one for now. Great working sticking this out!

lanklaas:
When will this be available in a release? Is there some workaround I can use in the current 0.9.0 release to close a tab?

PR #176, opened 2019-10-03
Easier access to launch options builder

atroche:
Useful (for me!) at least until this gets solved: https://www.bountysource.com/issues/49744518-add-generated-structs-as-autocomplete-suggestions


max-frai:
Looks like after this merge `LaunchOptions` are not available for external use (it's not defined in root crate and we can't access it directly from process)

PR #182, opened 2019-10-17
fetcher options bug fixes

zbidwell:
I finally got around to integrating this with my application and ran into a few things that weren't working correctly

PR #183, opened 2019-10-23
Give min version for zip (increase zip performance)

leshow:
The performance of `zip-rs` has historically not been that great, they switched recently to flate2 away from libflate, and it's been improved [a lot](https://github.com/mvdnes/zip-rs/pull/107). This PR just provides a min version number for `zip-rs`.

FYI, when the current master is run, it doesn't appear to download anything. It looks like there's another PR (#182) that gets everything working again, so I didn't spend much time with it.


atroche:
Ah, good to hear. Thanks!

Issue #184, opened 2019-11-15
Ignore SSL Errors


P3GLEG:
I was reading through the process.rs and wasn't able to find a way to add command line arguments. Probably for security reasons which I approve of. Is there any way you can add a field for `--ignore-certificate-errors`? I am doing some testing and require this flag in order to do local network testing not just localhost.

jeprojects:
@P3GLEG Added it so you can add any command line options to the args method of LaunchOptionsBuilder::default().

https://github.com/atroche/rust-headless-chrome/pull/195

P3GLEG:
Thanks man!

PR #185, opened 2019-11-15
Add LaunchOption to ignore SSL errors

P3GLEG:
Related to #184


atroche:
Would you mind changing the name from `verify` to `ignore_certificate_errors`? Just so it matches the CLI args more closely (and verify could refer to a number of things).

PR #191, opened 2019-11-26
Expose the ConnectionClosed Error

hengchu:
I have an application that uses this error, do you mind making it public?


atroche:
Not at all, thanks for the PR and sorry for the slow response :)

PR #204, opened 2020-02-18
Add --disable-software-rasterizer to DEFAULT_ARGS

Folyd:
In alpine docker container, without this flag, chromium will raise such an error: 
```
[0218/102856.761494:ERROR:gl_implementation.cc(282)] Failed to load /usr/lib/chromium/swiftshader/libGLESv2.so: Error loading shared library /usr/lib/chromium/swiftshader/libGLESv2.so: No such file or dir
ectory
```


Folyd:
This is an unnecessary PR, so I'll close it.

PR #195, opened 2019-12-24
LaunchOptions, Fetch Domain, authenticate()

jeprojects:
Added the ability to include command line args.
Example: LaunchOptionsBuilder::default().args(vec!["--proxy-server=127.0.0.1:8888"])

Added Fetch domain, with methods: enable_fetch(), disable_fetch(), enable_request_interception(), authenticate()

Added network::SetExtraHttpHeaders protocol and set_extra_http_headers()
Added network::SetCookies protocol and set_cookies() method

PR #206, opened 2020-03-05
Added extra_args for LaunchOptions

Folyd:


Folyd:
Duplicated with #195 . Sorry for that.

PR #209, opened 2020-03-15
Type non-latin text + insert text command

max-frai:
Allow to type non-latin text. Added method for inserting text into focused element on page.


max-frai:
Sorry, I accidently pushed non-related commit to the pull request title (stop_loading method). And my git knowledge is not enough to fix everything :)

PR #219, opened 2020-08-19
Allow You to Specify Extra Chrome Args

zicklag:
I'm experimenting with using this as an alternative backend to [Tauri](https://tauri.studio) and this PR adds the ability to customize chrome launch a little bit.

- Allow you to specify extra arbitrary chrome args
- Allow you to opt-out of the default chrome args
- Allow you to set an "app mode" URL with the LaunchOptionsBuilder
- Add an example that makes Chrome work kind of like Electron

### DIFF
--- a/examples/electron_like.rs
+++ b/examples/electron_like.rs
@@ -1,0 +1,23 @@
+use failure::Fallible;
+
+use headless_chrome::{Browser, LaunchOptionsBuilder};
+
+// This could possibly be used to integrate Rust as a way to create web-based desktop applications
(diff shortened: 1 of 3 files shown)


zicklag:
Looks like CI is failing because of the pedantic clippy lints, but those are not from things that I changed in this PR.

Billy-Sheppard:
Hi Zicklag

I know this PR is very old, and I believe some or all of your points have been addressed in newer versions, I am going to close this PR for now but if you wish to re-open it please let me know.

Issue #203, opened 2020-02-06
an example won't compile


GildedHonour:
the wikipedia example.
crate taken from `crates.io`

```

error[E0432]: unresolved import `headless_chrome::LaunchOptions`
 --> src/main.rs:5:66
  |
5 | use headless_chrome::{protocol::page::ScreenshotFormat, Browser, LaunchOptions};
  |                                                                  ^^^^^^^^^^^^^ no `LaunchOptions` in the root


```

GildedHonour:
```
error[E0599]: no method named `wait_for_initial_tab` found for type `std::result::Result<headless_chrome::browser::Browser, failure::error::Error>` in the current scope
  --> src/main.rs:20:23
   |
20 |     let tab = browser.wait_for_initial_tab();
   |                       ^^^^^^^^^^^^^^^^^^^^ method not found in `std::result::Result<headless_chrome::browser::Browser, failure::error::Error>`


```

Adeleet:
Can you link to the specific example? The [example from crates.io](https://docs.rs/headless_chrome/0.9.0/headless_chrome/#quick-start) is based on version 0.9.0 and does not use [`LaunchOptions` but uses the `LaunchOptionsBuilder` instead](https://docs.rs/headless_chrome/0.9.0/headless_chrome/struct.LaunchOptionsBuilder.html). If you mean [this example](https://github.com/atroche/rust-headless-chrome/blob/master/examples/query_wikipedia.rs) on this repo, the `LaunchOptions` was added to the crate-level namespace in #176 (after 0.9.0) so does not work with the version from crates.io. Until a new (stable) version is released on crates.io, you'll have to use `LaunchOptionsBuilder` or one of [these methods](https://docs.rs/headless_chrome/0.9.0/headless_chrome/browser/struct.Browser.html#method.new_tab_with_options) to launch with specific options.

PR #220, opened 2020-09-28
LaunchOptionsBuilder in examples

awikman:
Update the examples to use LaunchOptionsBuilder instead of LaunchOptions in the Browser constructor.
As discussed in issue #203


DCjanus:
Ping

PR #230, opened 2021-02-01
add print related launch options

FreeMasen:
I'd like to disable the print confirmation when running chrome via this crate. Exposing these config flags seems to do the trick. The one caveat here is that they seem to not have an effect unless `headless(false)` is also included in the options


Billy-Sheppard:
This functionality is more effectively reached by leveraging custom arguments with `LaunchOptionsBuilder::default().args()`. I am going to close this PR for now but if you'd like me to reopen it please let me know.

PR #233, opened 2021-03-09
added css marker pseudo element

juma121:
Added the css ::marker pseudo element that resulted tab.find_element() to panic when it present in the html.


atroche:
@juma121 thanks!

Issue #142, opened 2019-07-29
RPC connection times out after a while


jonas-schievink:
When I navigate to a tab and then do nothing for a couple of seconds while keeping the `Browser` instance around, eventually an error like the following will be logged:

```
Got a timeout while listening for browser events (Chrome #Some(28734))
```

Is there a way to avoid this, other than restarting the whole browser after this happens? I haven't found a way to configure timeouts in the documentation.

atroche:
Ah, sorry about that. I'll either make it configurable, or just not time out in the first place -- I can't remember why I started doing that in the first place.

In the meantime as a workaround you could do a no-op like querying the DOM every so often.

On Tue, 30 Jul 2019, at 7:41 AM, Jonas Schievink wrote:
> When I navigate to a tab and then do nothing for a couple of seconds while keeping the `Browser` instance around, eventually an error like the following will be logged:

> `Got a timeout while listening for browser events (Chrome #Some(28734))
`

jonas-schievink:
Hmm, the same also seems to happen when the `Browser` is dropped and I try to use a `Tab` from that browser instance.

atroche:
That same error message? Strange.

Do you have steps to reproduce?

jonas-schievink:
This is enough to trigger the second case:

```rust
fn main() {
    env_logger::init();

    let browser = headless_chrome::Browser::default().unwrap();
    let tab = browser.new_tab().unwrap();
    drop(browser);

    tab.navigate_to("https://google.com").unwrap();
}
```

This panics with:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ConnectionClosed

stack backtrace:
   0: failure::backtrace::internal::InternalBacktrace::new::heb1fa3adca37e9da (0x5574bf67bf23)
             at /home/jonas/.cargo/registry/src/github.com-1ecc6299db9ec823/failure-0.1.5/src/backtrace/internal.rs:44
   1: failure::backtrace::Backtrace::new::h4d905dfc1daea1c8 (0x5574bf67bcce)
             at /home/jonas/.cargo/registry/src/github.com-1ecc6299db9ec823/failure-0.1.5/src/backtrace/mod.rs:111
   2: <failure::error::error_impl::ErrorImpl as core::convert::From<F>>::from::hcd32f4b79305d2bd (0x5574bf16e017)
[...]

PR #237, opened 2021-04-07
Make a connect variant which accepts idle_browser_timeout

djozis:
This is a follow-up for issue #142, extending the solution provided in PR #145.


djozis:
@atroche does this look good to you? Just a simple consistency fix which otherwise blocks my use-case.

atroche:
Yep! Apologies for the slow response

Issue #236, opened 2021-04-02
Point is not constructable from the outside


DanielMSchmidt:
I have a page that heavily uses the shadow dom API, therefore I need to rely on relative positions or completely absolute ones for interacting with elements. I need to press a button in the bottom left corner and wanted to express sth like:

```rust
let bounds = tab.get_bounds()?;
let x = 80;
let y = bounds.height - 80;
// vvv module `point` is private
let point = headless_chrome::browser::tab::point::Point { x: x, y: y };
tab.click_point(point);
```

Am I missing a way to construct the point (I am pretty new to rust) or should this be public?

PR #238, opened 2021-04-14
Make LaunchOptions customizable and enable extracting midpoints after failure

mdrokz:
# Changes

https://github.com/atroche/rust-headless-chrome/commit/a93a0e2ac89da641ad73850115459bc8e22e8789 - i made LaunchOptions public and added access of tab in event closure

https://github.com/atroche/rust-headless-chrome/commit/a14d61d48a53912b0a66bafff5ca831c238fb66c - on low speed networks sometimes the dom renders very late, because of this it cant find midpoints to click on the element. so i added functionality to extract midpoints after failure.

### DIFF
--- a/examples/settings_and_event.rs
+++ b/examples/settings_and_event.rs
@@ -1,0 +1,42 @@
+use std::sync::Arc;
+
+use failure::Fallible;
+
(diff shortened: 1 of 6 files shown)


mdrokz:
Linking this issue - https://github.com/atroche/rust-headless-chrome/issues/236 as my commit makes the point module public

PR #239, opened 2021-04-20
Fetch element by xpath

mdrokz:
# Changes

https://github.com/mdrokz/rust-headless-chrome/commit/82c4bb9664d96db7e20ef6165a0c6181898fa44d - implemented feature to find elements by xpath and created a generic function that takes an enum to find element by either css selector or xpath query.

### DIFF
--- a/src/browser/tab/element/mod.rs
+++ b/src/browser/tab/element/mod.rs
@@ -1,7 +1,7 @@
 use std::fmt::Debug;
 use std::time::Duration;
 
-use failure::{Fail, Fallible};
+use failure::{Error, Fail, Fallible};
 use log::*;
 
 use crate::browser::tab::point::Point;
@@ -107,6 +107,30 @@ impl<'a> Element<'a> {
             .run_query_selector_on_node(self.node_id, selector)
     }
 
+    pub fn find_element_by_xpath(&self, query: &str) -> Fallible<Element<'_>> {
+        self.parent.get_document()?;
+
+        self.parent
+            .call_method(dom::methods::PerformSearch { query })
+            .and_then(|o| {
+                Ok(self
+                    .parent
+                    .call_method(dom::methods::GetSearchResults {
+                        search_id: &o.search_id,
+                        from_index: 0,
+                        to_index: o.result_count,
+                    })?
+                    .node_ids[0])
+            })
+            .and_then(|id| {
+                if id == 0 {
+                    Err(NoElementFound {}.into())
+                } else {
+                    Ok(Element::new(self.parent, id)?)
+                }
+            })
+    }
+
     /// Returns the first element in the document which matches the given CSS selector.
     ///
     /// Equivalent to the following JS:
@@ -141,6 +165,75 @@ impl<'a> Element<'a> {
     pub fn find_elements(&self, selector: &str) -> Fallible<Vec<Self>> {
         self.parent
             .run_query_selector_all_on_node(self.node_id, selector)
+    }
+
+    pub fn find_elements_by_xpath(&self, query: &str) -> Fallible<Vec<Element<'_>>> {
+        self.parent
+            .call_method(dom::methods::PerformSearch { query })
+            .and_then(|o| {
+                Ok(self
+                    .parent
+                    .call_method(dom::methods::GetSearchResults {
+                        search_id: &o.search_id,
+                        from_index: 0,
+                        to_index: o.result_count,
+                    })?
+                    .node_ids)
+            })
+            .and_then(|ids| {
+                ids.iter()
+                    .filter(|id| **id != 0)
+                    .map(|id| Element::new(self.parent, *id))
+                    .collect()
+            })
+    }
+
+    pub fn wait_for_element(&self, selector: &str) -> Fallible<Element<'_>> {
+        self.wait_for_element_with_custom_timeout(selector, Duration::from_secs(3))
+    }
+
+    pub fn wait_for_xpath(&self, selector: &str) -> Fallible<Element<'_>> {
+        self.wait_for_xpath_with_custom_timeout(selector, Duration::from_secs(3))
+    }
+
+    pub fn wait_for_element_with_custom_timeout(
+        &self,
+        selector: &str,
+        timeout: std::time::Duration,
+    ) -> Fallible<Element<'_>> {
+        debug!("Waiting for element with selector: {:?}", selector);
+        util::Wait::with_timeout(timeout).strict_until(
+            || self.find_element(selector),
+            Error::downcast::<NoElementFound>,
+        )
+    }
+
+    pub fn wait_for_xpath_with_custom_timeout(
+        &self,
+        selector: &str,
+        timeout: std::time::Duration,
+    ) -> Fallible<Element<'_>> {
+        debug!("Waiting for element with selector: {:?}", selector);
+        util::Wait::with_timeout(timeout).strict_until(
+            || self.find_element_by_xpath(selector),
+            Error::downcast::<NoElementFound>,
+        )
+    }
+
+    pub fn wait_for_elements(&self, selector: &str) -> Fallible<Vec<Element<'_>>> {
+        debug!("Waiting for element with selector: {:?}", selector);
+        util::Wait::with_timeout(Duration::from_secs(3)).strict_until(
+            || self.find_elements(selector),
+            Error::downcast::<NoElementFound>,
+        )
+    }
+
+    pub fn wait_for_elements_by_xpath(&self, selector: &str) -> Fallible<Vec<Element<'_>>> {
+        debug!("Waiting for element with selector: {:?}", selector);
+        util::Wait::with_timeout(Duration::from_secs(3)).strict_until(
+            || self.find_elements_by_xpath(selector),
+            Error::downcast::<NoElementFound>,
+        )
     }
 
     /// Moves the mouse to the middle of this element
--- a/src/browser/tab/element/mod.rs
+++ b/src/browser/tab/element/mod.rs
@@ -1,5 +1,5 @@
-use std::collections::HashMap;
 use std::fmt::Debug;
+use std::time::Duration;
 
 use failure::{Fail, Fallible};
 use log::*;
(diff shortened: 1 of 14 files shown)


src/browser/tab/element/mod.rs
```diff
@@ -337,38 +341,65 @@ impl<'a> Element<'a> {
...
+            }) {
+            Ok(e) => return Ok(e),
+            Err(_) => {
+                let mut p = Point { x: 0.0, y: 0.0 };
+
+                util::Wait::with_sleep(Duration::from_secs(1)).run_until(|| {
+                    let r = self.call_js_fn(
+                        r#"
+                    function() {
+                        let v = document.getElementsByClassName("ShowAllChapters")[0];
```
atroche, line 365: ShowAllChapters?

src/browser/tab/mod.rs
```diff
@@ -382,31 +400,31 @@ impl<'a> Tab {
         sleep(Duration::from_millis(scaled_millis));
     }
 
-    pub fn wait_for_element(&self, selector: &str) -> Fallible<Element<'_>> {
+    pub fn wait_for_element(&self, selector: Selector) -> Fallible<Element<'_>> {
```
atroche, line 403: It'd be nice to not have to break backwards compatibility here. Could you add a separate method for xpath, [like how Puppeteer has a separate method](https://pptr.dev/#?product=Puppeteer&version=v8.0.0&show=api-pagewaitforxpathxpath-options)?

atroche: Or use `Into` and make the default behavour for `"#asdfad".into()` be that it becomes a CSS selector?

src/util.rs
```diff
@@ -70,6 +93,20 @@ impl Wait {
         }
     }
 
+    pub fn run_until<F>(&self, predicate: F)
```
atroche, line 96: why do you need this method? in the case where you've used it, you can have the predicate return `Some(())` or `None` instead of true / false, right?

src/util.rs
```diff
@@ -70,6 +93,20 @@ impl Wait {
...
+    where
+        F: FnMut() -> bool,
+    {
+        let mut predicate = predicate;
+
+        loop {
+            if predicate() {
```
atroche, line 103: without using the timeout this could hang forever

src/browser/tab/mod.rs
```diff
@@ -58,6 +58,12 @@ pub type ResponseHandler = Box<
     + Sync,
 >;
 
+pub struct SyncSendEvent(pub Arc<Tab>, pub Box<dyn Fn(&Event, &Tab)>);
```
atroche, line 61: is this so that the event listener can have a reference to the `Tab`? why can't you just clone the tab and `move` it into the event listener closure?

atroche (changes requested):
In summary:

1. I don't think we should change the type of `SyncSendEvent` unless we really need to.
2. Let's keep `find_element` and its friends based on CSS selectors by default, to keep consistency with Puppeteer and also to keep the most common use case (which is definitely CSS selectors) ergonomic
3. The extra `run_until` method seems unnecessary and also dangerous (without timeout)

src/browser/tab/element/mod.rs
mdrokz: Sorrr this is embarrasing, i forgot to remove this

src/browser/tab/mod.rs
mdrokz: Originally i had a seperate method,but then i decided to follow what fantoccini did and made a function that accepted a enum, i will make a seperate method then,

src/browser/tab/mod.rs
mdrokz: yeah i previously tried to move the tab into the closure. but that wasnt working, so i changed the SyncSendEvent to accept the tab struct

src/util.rs
mdrokz: i use this method to constantly check for midpoints after failure. when the dom doesnt load sometimes the midpoint extraction fails, i previously tried using a timeout, i will change the code to use a custom timeout.

src/browser/tab/mod.rs
atroche: the reason i'm pushing back is that it'll break code that uses the current API, and I think you can get it working with moving the tab (maybe you need to make a new clone and move that?)

src/util.rs
atroche: but how come you can't just use the existing `until` method?

src/browser/tab/mod.rs
mdrokz: Yeah i understand, this is my first time contributing, so im a bit inexperienced, i will try doing that. i need some guidance 😅

src/util.rs
mdrokz: i did use that before but it required me to return a result,which i felt unnecessary, since i could return true or false.

src/browser/tab/mod.rs
atroche: no worries, thanks for taking the time to respond to my feedback 🙂

src/util.rs
mdrokz: i wanted to make this method for specific rare use cases such as on a slow internet the dom can render very late. but your right

src/browser/tab/mod.rs
mdrokz: > Or use `Into` and make the default behavour for `"#asdfad".into()` be that it becomes a CSS selector?

this seems like a nice idea,  i will try doing that

src/browser/tab/mod.rs
mdrokz: > no worries, thanks for taking the time to respond to my feedback

let me try again with cloning the tab and moving it in the closure.

src/browser/tab/element/mod.rs
mdrokz: > Sorry this is embarrasing, i forgot to remove this

i was testing it with a element and forgot to remove it in my fork

src/util.rs
atroche: In your example you do:

```
p = v;
true
```

where p is a mut variable defined above the call to `run_until`. but you could instead do `p = until(.....)` and have the until return `Some(v)`, which will become `p`.

mdrokz: > In your example you do:
> 
> ```
> p = v;
> true
> ```
> 
> where p is a mut variable defined above the call to `run_until`. but you could instead do `p = until(.....)` and have the until return `Some(v)`, which will become `p`.

I see, that definitely seems a better way.

src/browser/tab/mod.rs
atroche: cool, and if you call `into()` inside `find_element` then people can still call `find_element('#example')` (i.e. it won't change the public API)

mdrokz: > cool, and if you call `into()` inside `find_element` then people can still call `find_element('#example')` (i.e. it won't change the public API)

Got it

src/browser/tab/mod.rs
mdrokz: I tried again with cloning the tab and it was working, previously i didnt clone it properly.

mdrokz:
> In summary:
> 
> 1. I don't think we should change the type of `SyncSendEvent` unless we really need to.
> 2. Let's keep `find_element` and its friends based on CSS selectors by default, to keep consistency with Puppeteer and also to keep the most common use case (which is definitely CSS selectors) ergonomic
> 3. The extra `run_until` method seems unnecessary and also dangerous (without timeout)

I fixed all of the issues you have mentioned in this commit - https://github.com/atroche/rust-headless-chrome/pull/239/commits/b68cfa53051d625c5fb773962688cc200627b31f

atroche:
@mdrokz nice one!

With [this](https://github.com/atroche/rust-headless-chrome/commit/b68cfa53051d625c5fb773962688cc200627b31f#diff-d91ca1f76b204ad2cf6d1781c3fbbd7559d0dd60b553dac372772308c14db558R79-R90), I think there are ways that can go wrong (e.g. if you use an XPath expression that contains a `.` or a `#` in it).

Instead, could you make strings always convert to a `Selector::CSS`, and make `wait_for_element` take something that satisfies `Into<Selector>`?

That way they can pass in a `str` or a `Selector::XPath`. Otherwise we might have to just add separate methods for finding via XPath.

mdrokz:
> @mdrokz nice one!
> 
> With [this](https://github.com/atroche/rust-headless-chrome/commit/b68cfa53051d625c5fb773962688cc200627b31f#diff-d91ca1f76b204ad2cf6d1781c3fbbd7559d0dd60b553dac372772308c14db558R79-R90), I think there are ways that can go wrong (e.g. if you use an XPath expression that contains a `.` or a `#` in it).
> 
> Instead, could you make strings always convert to a `Selector::CSS`, and make `wait_for_element` take something that satisfies `Into<Selector>`?
> 
> That way they can pass in a `str` or a `Selector::XPath`. Otherwise we might have to just add separate methods for finding via XPath.

I see, i will try implementing what you suggested.

mdrokz:
> > @mdrokz nice one!
> > With [this](https://github.com/atroche/rust-headless-chrome/commit/b68cfa53051d625c5fb773962688cc200627b31f#diff-d91ca1f76b204ad2cf6d1781c3fbbd7559d0dd60b553dac372772308c14db558R79-R90), I think there are ways that can go wrong (e.g. if you use an XPath expression that contains a `.` or a `#` in it).
> > Instead, could you make strings always convert to a `Selector::CSS`, and make `wait_for_element` take something that satisfies `Into<Selector>`?
> > That way they can pass in a `str` or a `Selector::XPath`. Otherwise we might have to just add separate methods for finding via XPath.
> 
> I see, i will try implementing what you suggested.

On another thought, i think it might better to just implement seperate methods.

mdrokz:
I made the xpath methods seperate in this commit - https://github.com/atroche/rust-headless-chrome/commit/35113dbc5837d2ebdea69d41c4e2c579a17799c2

mdrokz:
Is this ready for merge now?

PR #242, opened 2021-05-11
Fix build

k-bx:
Initially intended to add the set-cookie functionality, but I've since found that the code is present, just the repo doesn't really build, so the changes are just fixing build errors

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ Cargo.lock
 !/tests/temp/.gitkeep
 
 debug.log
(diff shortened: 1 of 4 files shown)


mdrokz:
> Initially intended to add the set-cookie functionality, but I've since found that the code is present, just the repo doesn't really build, so the changes are just fixing build errors

Thanks, actually i merged a different branch to master and it overrode the previous changes that i did and thats why it had errors

PR #249, opened 2021-06-01
Remove unused imports to pass clippy tests

striezel:
This PR removes some unused imports and prefixes an unused variable with an underscore. Purpose of these changes is to get rid of the associated compiler warnings.


striezel:
Travis CI builds still throw errors (like on the master branch (https://travis-ci.com/github/atroche/rust-headless-chrome/jobs/506590317)), but now they get past the `clippy` checks (https://travis-ci.com/github/atroche/rust-headless-chrome/jobs/509865081) and actually run tests.

PR #275, opened 2021-10-28
fix ignore_certificate_errors flag, idle_browser_timeout, process stderr handle

initprism:
The logic to enable the option was invalid.


mdrokz:
Thank you for your contribution.

PR #293, opened 2022-02-14
Browser sync

### LINKED ISSUE #291: Cloning Browser and using a mpsc::SyncSender
I'd like to suggest `impl Clone on Browser` as well as using a `SyncSender` for the `mpsc` channel.

This will allow the browser to be cheaply clonable and used when Sync is needed.

My current use case is generating PDFs for routes using warp server, however, right now I need to create a browser for each request in the route. Even if I wrap the instance in `Arc<Browser>` it is still `!Sync` and cannot be shared between threads

elpiel:
Closes #291 

Also fixes the fetcher module (feature = "fetch"), as `failure` doesn't seem to be used anymore + a few other problem that came up.

### DIFF
--- a/src/browser/transport/mod.rs
+++ b/src/browser/transport/mod.rs
@@ -61,7 +61,7 @@ pub struct Transport {
     listeners: Listeners,
     open: Arc<AtomicBool>,
     call_id_counter: Arc<AtomicU32>,
-    loop_shutdown_tx: Mutex<mpsc::Sender<()>>,
+    loop_shutdown_tx: Mutex<mpsc::SyncSender<()>>,
 }
 
 #[derive(Debug, Error)]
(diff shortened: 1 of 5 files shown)


src/browser/transport/mod.rs
```diff
@@ -84,7 +84,7 @@ impl Transport {
 
         let open = Arc::new(AtomicBool::new(true));
 
-        let (shutdown_tx, shutdown_rx) = mpsc::channel();
+        let (shutdown_tx, shutdown_rx) = mpsc::sync_channel(1);
```
mdrokz, line 87: Can you take a larger buffer value here ? like 50 or 100, a small buffer size can impact peformance

mdrokz (commented):
Everything looks good, thanks for your contribution.

elpiel:
I've changed the buffer size for both channels.

There is also a test I can commit that checks that Browser is Sync

src/browser/transport/mod.rs
elpiel: Fixed.

PR #297, opened 2022-02-26
add MIT LICENSE.md file

### LINKED ISSUE #282: Add license file
I see the [`Cargo.toml`](https://github.com/atroche/rust-headless-chrome/blob/09747c06c77a305adf2421d67d74ae69c0430225/Cargo.toml#L7) indicates this project is licensed under the MIT license, but I couldn't find a license file.

Would it be possible to add a license file that includes the copyright required by the MIT license (https://choosealicense.com/licenses/mit/)?

elpiel:
Closes #282

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,7 @@
+Copyright © 2022 headless_chrome contributors
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the “Software”), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+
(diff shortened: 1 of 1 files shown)


LICENSE.md
```diff
@@ -0,0 +1,7 @@
+Copyright © 2022 headless_chrome contributors
```
elpiel, line 1: The only line I'm not 100% if it should stay like this as this should be the `<copyright holders>`
If you have better suggestions I'm open to them.

elpiel:
It is stated in Cargo manifest docs that:

> If a package is using a nonstandard license, then the license-file field may be specified in lieu of the license field.

https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields

Since MIT is a standard license, I don't think there is a need for `license_file`.

mdrokz:
> It is stated in Cargo manifest docs that:
> 
> > If a package is using a nonstandard license, then the license-file field may be specified in lieu of the license field.
> 
> https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields
> 
> Since MIT is a standard license, I don't think there is a need for `license_file`.

I see, i will merge this PR then thanks for your contribution

PR #309, opened 2022-06-16
Fix ResourceType naming for XHR and CSPViolationReport

Oyami-Srk:
XHR and CSPViolationReport are not strict PascalCase. Should add a special case renaming to avoid failure.

Renaming enum to `XHR` and `CSPViolationReport` seems more consistent. I think it would be "backward-compatible" since there is no one reporting this bug : /

[ResourceType on ChromeDev Protocol manual](https://chromedevtools.github.io/devtools-protocol/tot/Network/#type-ResourceType),

### DIFF
--- a/src/protocol.rs
+++ b/src/protocol.rs
@@ -9244,6 +9244,7 @@ pub mod cdp {
             Font,
             Script,
             TextTrack,
+            #[serde(rename = "XHR")]
             Xhr,
             Fetch,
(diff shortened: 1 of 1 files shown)


Oyami-Srk:
Since the generator side fixing is merged (https://github.com/mdrokz/auto_generate_cdp/pull/1)

Close this.

PR #310, opened 2022-06-17
Created `send_character` method as a much faster alternative to `type_str`

DanielPikilidis:
In addition to `type`, puppeteer also has this method for text input: [`sendCharacter`](https://github.com/puppeteer/puppeteer/blob/ce0dd25349b3141409cf68121a2bf9a770d9ecf7/src/common/Input.ts#L243)

What this does is it just passes the provided input text to the `Input::InsertText` method without any processing. This is very useful because when we have a lot of text, the normal `type` (or `type_str` in our case) method takes a very long time. The `sendCharacter` method does it instantly.

### Changes

1. Created the `send_character` method.
2. Changed `type_str` so now it uses the new `send_character` method instead of calling `Input::InsertText` directly. No real reason for this change, but I think it makes the code a little more organized.

<br>

_Note: I'm not very happy with neither the method name "send_character" nor the parameter name "char_to_send". We should definitely find better names because even thought there is documentation, I believe the method name should give to the developer some idea of what it does._

This is the documentation I made: (copied from puppeteer docs)
Screenshot from 2022-06-17 16-21-51

### DIFF
--- a/src/browser/tab/mod.rs
+++ b/src/browser/tab/mod.rs
@@ -801,10 +801,11 @@ impl Tab {
         Ok(self)
     }
 
-    /// Dispatches a `keypress` and `input` event.
-    /// This does not send a `keydown` or `keyup` event.
+    /// Does the same as `type_str` but it only dispatches a `keypress` and `input` event.
+    /// It does not send a `keydown` or `keyup` event.
     ///
-    /// Useful when you have a lot of text to input. With this method it is instant
+    /// What this means is that it is much faster. 
+    /// It is especially useful when you have a lot of text as input.
     pub fn send_character(&self, char_to_send: &str) -> Result<&Self> {
         self.call_method(Input::InsertText {
             text: char_to_send.to_string(),
--- a/src/browser/tab/mod.rs
+++ b/src/browser/tab/mod.rs
@@ -794,12 +794,21 @@ impl Tab {
                     })?;
                 }
                 Err(_) => {
-                    self.call_method(Input::InsertText {
-                        text: c.to_string(),
-                    })?;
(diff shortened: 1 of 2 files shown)


mdrokz:
> In addition to `type`, puppeteer also has this method for text input: [`sendCharacter`](https://github.com/puppeteer/puppeteer/blob/ce0dd25349b3141409cf68121a2bf9a770d9ecf7/src/common/Input.ts#L243)
> 
> What this does is it just passes the provided input text to the `Input::InsertText` method without any processing. This is very useful because when we have a lot of text, the normal `type` (or `type_str` in our case) method takes a very long time. The `sendCharacter` method does it instantly.
> 
> ### Changes
> 
> 1. Created the `send_character` method.
> 2. Changed `type_str` so now it uses the new `send_character` method instead of calling `Input::InsertText` directly. No real reason for this change, but I think it makes the code a little more organized.
> 
> <br>
> 
> _Note: I'm not very happy with neither the method name "send_character" nor the parameter name "char_to_send". We should definitely find better names because even thought there is documentation, I believe the method name should give to the developer some idea of what it does._
> 
> This is the documentation I made: (copied from puppeteer docs)
> Screenshot from 2022-06-17 16-21-51

Nice work & thanks for your contribution before i merge can you write a test for this functionality?

DanielPikilidis:
The test is pretty much the same as the one above it (form_interaction) but instead of using type_into on the element I'm using send_character after selecting the element.

mdrokz:
> The test is pretty much the same as the one above it (form_interaction) but instead of using type_into on the element I'm using send_character after selecting the element.

very good thanks, i will merge this now

Issue #311, opened 2022-06-20
temporary profiles are not cleaned up automatically


djfm:
I have a crawler running for a couple of days, it spawns a new browser on each URL, so it creates many instances... It took me some time to figure out why it ended up crashing: my `/tmp` dir is saturated with `rust-headless-chrome-profile3kcd4a` etc., haven't have time to investigate much yet, will be today's goal.

PR #312, opened 2022-06-20
Automatically delete temporary user_data_dir when none was explicitly provided by the user

djfm:
Currently it seems that the chrome profiles that are automatically created when spawning a browser without specifying the user_data_dir are not deleted, resulting in high disk space usage.

The bug seems to happen because the instance of `tempfile::TempDir`, which handles deletion of the directory upon drop, is not kept around after the browser is launched (see https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html#resource-leaking).
    
This should fix the issue, reported here:
https://github.com/atroche/rust-headless-chrome/issues/311

I'm not very happy about the (minimal) use of `unsafe` code in the test code and I'm open to suggestions.
I opted for the dirty global static solution for testing because I did not want to alter the lib's API and expose that temporary path too high up, but other solutions could be thought of if this hurts your eyes too much.

### DIFF
--- a/src/browser/process.rs
+++ b/src/browser/process.rs
@@ -62,7 +62,7 @@ impl Drop for TemporaryProcess {
 /// binary on the system, use an available port for debugging, and start in headless mode.
 #[derive(Builder)]
 pub struct LaunchOptions<'a> {
-    /// Determintes whether to run headless version of the browser. Defaults to true.
+    /// Determines whether to run headless version of the browser. Defaults to true.
     #[builder(default = "true")]
     pub headless: bool,
 
@@ -112,7 +112,7 @@ pub struct LaunchOptions<'a> {
 
     /// Disable default arguments
     #[builder(default)]
-    pub disable_default_args: bool,    
+    pub disable_default_args: bool,
 
     /// The options to use for fetching a version of chrome when `path` is None.
     ///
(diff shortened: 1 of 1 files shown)


mdrokz:
> Currently it seems that the chrome profiles that are automatically created when spawning a browser without specifying the user_data_dir are not deleted, resulting in high disk space usage.
> 
> The bug seems to happen because the instance of `tempfile::TempDir`, which handles deletion of the directory upon drop, is not kept around after the browser is launched (see https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html#resource-leaking).
> 
> This should fix the issue, reported here: #311
> 
> I'm not very happy about the (minimal) use of `unsafe` code in the test code and I'm open to suggestions. I opted for the dirty global static solution for testing because I did not want to alter the lib's API and expose that temporary path too high up, but other solutions could be thought of if this hurts your eyes too much.

Hey thanks for your contribution, everything looks good & the unsafe use is fine for tests i will merge this now.

djfm:
Btw you don't have CI on PR's?

mdrokz:
> Btw you don't have CI on PR's?

it used to have travis CI, now the owner disabled it and i dont have permissions.

djfm:
Ah OK, now GirHub can run the CI it's much more powerful than Travis :)

PR #314, opened 2022-06-23
Add multiple response handlers and allow deregistration

Oyami-Srk:
Deregistration is useful for some conditions like reusing a tab for different websites OR the needing for inspecting one specified response and disabling response handle.

I'd like to use `HashSet` to implement this. But it's not easy to make `ResponseHandler` Hashable.


mdrokz:
> Deregistration is useful for some conditions like reusing a tab for different websites OR the needing for inspecting one specified response and disabling response handle.
> 
> I'd like to use `HashSet` to implement this. But it's not easy to make `ResponseHandler` Hashable.

sorry for late reply, thanks for your contribution i will merge this now.

PR #315, opened 2022-06-30
Created methods to get full HTML from page and elements

DanielPikilidis:
Adds two new methods for get the full HTML contents of a tab or a selected element.

The String returned by `tab.get_content()` can be parsed by [scraper](https://crates.io/crates/scraper) with the method `scraper::Html::parse_document`. 
The String returned by `elem.get_content()` can be parsed by [scraper](https://crates.io/crates/scraper) with the method `scraper::Html::parse_fragment`.

The way it gets the HTML is by running a JS function. I don't know if this is the best way to do it, but I can't think of any other way.

### DIFF
--- a/src/browser/tab/element/mod.rs
+++ b/src/browser/tab/element/mod.rs
@@ -352,6 +352,18 @@ impl<'a> Element<'a> {
         Ok(text)
     }
 
+    /// Get the full HTML contents of the element.
+    /// 
+    /// Equivalent to the following JS: ```element.outerHTML```.
+    pub fn get_content(&self) -> Result<String> {
+        let html = self.
+                call_js_fn("function() { return this.outerHTML }", vec![], false)?
+                    .value
+                    .unwrap();
+
+        Ok(String::from(html.as_str().unwrap()))
+    }
+
     pub fn get_computed_styles(&self) -> Result<Vec<CSSComputedStyleProperty>> {
         let styles = self
             .parent
--- a/src/browser/tab/mod.rs
+++ b/src/browser/tab/mod.rs
@@ -718,6 +718,17 @@ impl Tab {
             .root)
     }
 
+    /// Get the full HTML contents of the page.
+    /// 
+    /// Equivalent to the following JS: ```document.documentElement.outerHTML```
+    pub fn get_content(&self) -> Result<String> {
+        let html = self
+            .evaluate("(function () { return document.documentElement.outerHTML })();", false)?
+                .value
+                .unwrap();
+        Ok(String::from(html.as_str().unwrap()))
+    }
+
     pub fn find_elements(&self, selector: &str) -> Result<Vec<Element<'_>>> {
         trace!("Looking up elements via selector: {}", selector);
 
--- a/src/browser/tab/mod.rs
+++ b/src/browser/tab/mod.rs
@@ -719,11 +719,18 @@ impl Tab {
     }
 
     /// Get the full HTML contents of the page.
-    /// 
-    /// Equivalent to the following JS: ```document.documentElement.outerHTML```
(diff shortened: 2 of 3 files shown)


mdrokz:
https://chromedevtools.github.io/devtools-protocol/tot/DOM/#method-getOuterHTML can you try to use this ?

src/browser/tab/mod.rs
```diff
@@ -718,6 +718,17 @@ impl Tab {
...
+    /// Get the full HTML contents of the page.
+    /// 
+    /// Equivalent to the following JS: ```document.documentElement.outerHTML```
+    pub fn get_content(&self) -> Result<String> {
+        let html = self
+            .evaluate("(function () { return document.documentElement.outerHTML })();", false)?
+                .value
+                .unwrap();
+        Ok(String::from(html.as_str().unwrap()))
+    }
```
mdrokz, line 730: same with this https://chromedevtools.github.io/devtools-protocol/tot/DOM/#method-getOuterHTML

mdrokz (changes requested):

DanielPikilidis:
I tried using that but I never managed to get it to work. For some reason `DOM::getOuterHTML` always returns an empty string even though I'm passing the correct NodeId. If I put this code in `tab.get_content()` I always get back an empty string.
<pre>
let root_node_id = self.get_document()?.node_id;
let html = self
  .call_method(DOM::GetOuterHTML {
    node_id: Some(root_node_id),
    backend_node_id: None,
    object_id: None,
  })?.outer_html; 
</pre>

Same happens in the `element.get_content()` method.

mdrokz:
> I tried using that but I never managed to get it to work. For some reason `DOM::getOuterHTML` always returns an empty string even though I'm passing the correct NodeId. If I put this code in `tab.get_content()` I always get back an empty string.
> 
> let root_node_id = self.get_document()?.node_id;
> let html = self
>   .call_method(DOM::GetOuterHTML {
>     node_id: Some(root_node_id),
>     backend_node_id: None,
>     object_id: None,
>   })?.outer_html; 
> Same happens in the `element.get_content()` method.

hmm i checked the puppeteer code and it seems they arent using `getOuterHtml` API too

src/browser/tab/mod.rs
mdrokz: refactor the code to this ref https://github.com/puppeteer/puppeteer/blob/dd275590903368aab107912752b975f3b0a16489/src/common/DOMWorld.ts#L253-L264

src/browser/tab/element/mod.rs
```diff
@@ -352,6 +352,18 @@ impl<'a> Element<'a> {
         Ok(text)
     }
 
+    /// Get the full HTML contents of the element.
+    /// 
+    /// Equivalent to the following JS: ```element.outerHTML```.
+    pub fn get_content(&self) -> Result<String> {
+        let html = self.
+                call_js_fn("function() { return this.outerHTML }", vec![], false)?
```
mdrokz, line 360: can you change this too to the puppeteer version?

mdrokz (changes requested):

src/browser/tab/element/mod.rs
DanielPikilidis: I don't think puppeteer has a built-in method to get the HTML of a selected element, not one that I found at least. 
On a StackOverflow [question](https://stackoverflow.com/questions/46431288/puppeteer-get-innerhtml) about this they used the evaluate method.

mdrokz: ah i see thats ok then

mdrokz (approved):
looks good now, thank you for your contribution.

PR #320, opened 2022-08-01
Fix outdated code examples

qiaoruntao:
This pull request fixes two outdated code example. One incorrect markdown languange mark is also fixed.


mdrokz:
> This pull request fixes two outdated code example. One incorrect markdown languange mark is also fixed.

Thanks alot for your contribution.

PR #328, opened 2022-09-06
Added `microsoft-edge` to list of bins to look for

Billy-Sheppard:
Added another entry for microsoft edge to autodetect with.

The binary name for msedge on some OS's (some linux distros namely such as Silverblue) is `microsoft-edge`.


mdrokz:
> Added another entry for microsoft edge to autodetect with.
> 
> The binary name for msedge on some OS's (some linux distros namely such as Silverblue) is `microsoft-edge`.

Thank you for your contribution !

PR #338, opened 2022-11-04
Impl Deserialize and Default on PrintToPdfOptions

MortenLohne:
Thanks for a great library! I needed one small change for my use case, here it is if you want to merge it upstream. 

I'm using the library for a http microservice that accepts a html template _and_ `PrintToPdfOptions`, and generates a pdf. This needs a `Deserialize` impl to be possible, and also a `Default` impl to allow the field to be optional.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -73,7 +73,7 @@ impl Into<Option<Page::PrintToPDFTransfer_modeOption>> for TransferMode {
     }
 }
 
-#[derive(Deserialize, Serialize, Debug)]
+#[derive(Deserialize, Serialize, Debug, Default)]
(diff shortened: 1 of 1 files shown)


mdrokz (approved):

mdrokz:
> Thanks for a great library! I needed one small change for my use case, here it is if you want to merge it upstream.
> 
> I'm using the library for a http microservice that accepts a html template _and_ `PrintToPdfOptions`, and generates a pdf. This needs a `Deserialize` impl to be possible, and also a `Default` impl to allow the field to be optional.

Hi! Thanks for your contribution

PR #343, opened 2022-11-21
build: Add feature for enabling rustls in deps

seanpianka:
If any downstream crates are using rustls, they will potentially have issues since the current `auto_generate_cdp/reqwest` dependency uses the `native-tls` default feature. This change would allow for this to be toggled.

Requires https://github.com/mdrokz/auto_generate_cdp/pull/2 to be merged.

Signed-off-by: Sean Pianka <pianka@eml.cc>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,13 +40,14 @@ jpeg-decoder = { version = "0.1", default_features = false }
 filepath = "0.1.1"
 
 [build-dependencies]
-auto_generate_cdp = "0.3.3"
+auto_generate_cdp = { version = "0.3.3", default-features = false }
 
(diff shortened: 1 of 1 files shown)


mdrokz (approved):
I have approved & merged https://github.com/mdrokz/auto_generate_cdp/pull/2 and will be merging this now. thanks for your contribution !

seanpianka:
I left a note in https://github.com/mdrokz/auto_generate_cdp/pull/2#issuecomment-1324324104, but we'll need to bump the version of `auto_generate_cdp` from `0.3.3` as the `rustls` feature hasn't been released yet.

PR #348, opened 2022-12-22
Fix ftbts

allquixotic:
The current master branch fails to build from source when specifying my own serde / serde_json dependency in my project, unless I fix the code according to this PR. Looks like a Result needs to be unwrapped and an error propagated on these two lines in fetcher.rs.


Billy-Sheppard:
Looks good! Thanks for your work on this - I am unable to test this atm but I'll merge it in in the next day or two, unless @mdrokz gets to it first

mdrokz (approved):
Looks good, thanks for your contribution!

Issue #346, opened 2022-12-17
Move from websocket to tungstenite 


sinistersnare:
The [`websocket` crate](https://crates.io/crates/websocket) That this project depends on recommends moving away from it, to other projects such as [`tungstenite`](https://crates.io/crates/tungstenite).

Currrently, compiling this crate creates a warning from cargo, due to a transitive dependency on `traitobject 0.1.0`, which is using deprecated Rust features. This warning is directly due to the `websocket` dependency.

mdrokz:
> The [`websocket` crate](https://crates.io/crates/websocket) That this project depends on recommends moving away from it, to other projects such as [`tungstenite`](https://crates.io/crates/tungstenite).
> 
> Currrently, compiling this crate creates a warning from cargo, due to a transitive dependency on `traitobject 0.1.0`, which is using deprecated Rust features. This warning is directly due to the `websocket` dependency.

see #322 #341

sinistersnare:
Ah sorry, I mostly made this as something to close, cause I was going to look to replace this right now. Ill look to close #322 instead

Issue #341, opened 2022-11-17
Swap out websocket impl from websocket_rs to tungstenite?


seanpianka:
The `websocket` package has a note in its repository that:
> "Maintainership of this project is slugglish. You may want to use tungstenite or tokio-tungstenite instead."

Upon further inspection, there has been [an open issue](https://github.com/websockets-rs/rust-websocket/issues/262) that the required version of Hyper (0.10.x) is vulnerable to several security vulnerabilities: RUSTSEC-2021-0079, RUSTSEC-2021-0078

Other issues have mention that "rust-websocket is based on long-legacy components and porting it to modern stack may be close to a rewrite," with recommendations to switch to tokio-tungstenite or async-tungstenite. This would allow an update to a modern version of Hyper.

Anyone have an idea of the effort required to gut `websocket`?

mdrokz:
> The `websocket` package has a note in its repository that:
> 
> > "Maintainership of this project is slugglish. You may want to use tungstenite or tokio-tungstenite instead."
> 
> Upon further inspection, there has been [an open issue](https://github.com/websockets-rs/rust-websocket/issues/262) that the required version of Hyper (0.10.x) is vulnerable to several security vulnerabilities: RUSTSEC-2021-0079, RUSTSEC-2021-0078
> 
> Other issues have mention that "rust-websocket is based on long-legacy components and porting it to modern stack may be close to a rewrite," with recommendations to switch to tokio-tungstenite or async-tungstenite. This would allow an update to a modern version of Hyper.
> 
> Anyone have an idea of the effort required to gut `websocket`?

you can see this comment for more info https://github.com/atroche/rust-headless-chrome/issues/322#issuecomment-1220587676

mdrokz:
closing this now refer to #322

PR #349, opened 2022-12-25
security: Fix vulnerability by switching websocket library

### LINKED ISSUE #322: Adress vulnerabilities reported by `cargo audit`
Running `cargo audit` will result in following vulnerabilities listed: 

<details>
<summary>
Output from <tt>cargo audit</tt>
</summary>

```text
azatoth@azaboxen:~/build/rust-websocket «master u=»$ cargo audit
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 433 security advisories (from /home/azatoth/.cargo/advisory-db)
    Updating crates.io index
    Updating crates.io index
    Scanning Cargo.lock for vulnerabilities (132 crate dependencies)
Crate:     hyper
Version:   0.10.16
Title:     Integer overflow in `hyper`'s parsing of the `Transfer-Encoding` header leads to data loss
Date:      2021-07-07
ID:        RUSTSEC-2021-0079
URL:       https://rustsec.org/advisories/RUSTSEC-2021-0079
Solution:  Upgrade to >=0.14.10
Dependency tree:
hyper 0.10.16
└── websocket 0.26.5

Crate:     hyper
Version:   0.10.16
Title:     Lenient `hyper` header parsing of `Content-Length` could allow request smuggling
Date:      2021-07-07
ID:        RUSTSEC-2021-0078
URL:       https://rustsec.org/advisories/RUSTSEC-2021-0078
Solution:  Upgrade to >=0.14.10

Crate:     time
Version:   0.1.44
Title:     Potential segfault in the time crate
Date:      2020-11-18
ID:        RUSTSEC-2020-0071
URL:       https://rustsec.org/advisories/RUSTSEC-2020-0071
Solution:  Upgrade to >=0.2.23
Dependency tree:
time 0.1.44
└── hyper 0.10.16
    └── websocket 0.26.5

Crate:     tokio
Version:   0.1.22
Title:     Data race when sending and receiving after closing a `oneshot` channel
Date:      2021-11-16
ID:        RUSTSEC-2021-0124
URL:       https://rustsec.org/advisories/RUSTSEC-2021-0124
Solution:  Upgrade to >=1.8.4, <1.9.0 OR >=1.13.1
Dependency tree:
tokio 0.1.22
├── websocket-base 0.26.5
│   └── websocket 0.26.5
└── websocket 0.26.5

Crate:     net2
Version:   0.2.37
Warning:   unmaintained
Title:     `net2` crate has been deprecated; use `socket2` instead
Date:      2020-05-01
ID:        RUSTSEC-2020-0016
URL:       https://rustsec.org/advisories/RUSTSEC-2020-0016
Dependency tree:
net2 0.2.37
├── miow 0.2.2
│   └── mio 0.6.23
│       ├── tokio-tcp 0.1.4
│       │   ├── websocket-base 0.26.5
│       │   │   └── websocket 0.26.5
│       │   ├── websocket 0.26.5
│       │   └── tokio 0.1.22
│       │       ├── websocket-base 0.26.5
│       │       └── websocket 0.26.5
│       ├── tokio-reactor 0.1.12
│       │   ├── websocket 0.26.5
│       │   ├── tokio-tcp 0.1.4
│       │   └── tokio 0.1.22
│       └── tokio 0.1.22
└── mio 0.6.23

error: 4 vulnerabilities found!
warning: 1 allowed warning found
```

</details>

Most of the issues seems to be related to the use of [`websocket`](https://github.com/websockets-rs/rust-websocket), and looking over there, it's evident they have fallen back on the maintenance, and even recommends people using other websocket libraries.

Question then is if it's feasible for `headless-chrome` to switch from `websocket` to for example [`tungstenite`](https://crates.io/crates/tungstenite) or [`tokio-tungstenite`](https://crates.io/crates/tokio-tungstenite)

1c3t3a:
Currently this crate uses the `websocket` crate, that has a dependency on a hyper version that is subject to a security vulnerability. This change switches the implementation to the `tokio-tungstenite` websocket library, which exposes an async interface. In order to fit the async interface with the current sync interface, [async-bridging](https://tokio.rs/tokio/topics/bridging) is used.

This fixes #322. 
Related issues: #346, #341

### DIFF
--- a/src/browser/transport/web_socket_connection.rs
+++ b/src/browser/transport/web_socket_connection.rs
@@ -1,26 +1,25 @@
 use std::sync::mpsc;
-use std::sync::Mutex;
+use std::sync::Arc;
 
 use anyhow::Result;
-use log::{debug, info, trace, warn};
-use websocket::client::sync::Client;
-use websocket::stream::sync::TcpStream;
-use websocket::url::Url;
-use websocket::WebSocketError;
-use websocket::{ClientBuilder, OwnedMessage};
+use futures::{
+    stream::{SplitSink, SplitStream},
+    SinkExt,
+};
+use futures_util::stream::StreamExt;
+use log::*;
+use tokio::{net::TcpStream, runtime::Runtime, sync::Mutex};
+use tokio_tungstenite::{connect_async, MaybeTlsStream, WebSocketStream};
+use tungstenite::client::IntoClientRequest;
+use url::Url;
 
 use crate::types::{parse_raw_message, Message};
 
+#[derive(Debug)]
 pub struct WebSocketConnection {
-    sender: Mutex<websocket::sender::Writer<TcpStream>>,
+    sender: Arc<Mutex<SplitSink<WebSocketStream<MaybeTlsStream<TcpStream>>, tungstenite::Message>>>,
+    runtime: Arc<Runtime>,
     process_id: Option<u32>,
-}
-
-// TODO websocket::sender::Writer is not :Debug...
-impl std::fmt::Debug for WebSocketConnection {
-    fn fmt(&self, f: &mut std::fmt::Formatter) -> Result<(), std::fmt::Error> {
-        write!(f, "WebSocketConnection {{}}")
-    }
 }
 
 impl WebSocketConnection {
@@ -29,17 +28,36 @@ impl WebSocketConnection {
         process_id: Option<u32>,
         messages_tx: mpsc::Sender<Message>,
     ) -> Result<Self> {
-        let connection = Self::websocket_connection(ws_url)?;
-        let (websocket_receiver, sender) = connection.split()?;
+        let runtime = Arc::new(
+            tokio::runtime::Builder::new_current_thread()
+                .enable_all()
+                .build()?,
+        );
 
+        let connection: Result<_> = runtime.block_on(async {
+            let (stream, _) = connect_async(ws_url.to_string().into_client_request()?).await?;
+
+            debug!("Successfully connected to WebSocket: {}", ws_url);
+
+            Ok(stream)
+        });
+
+        let (sender, recv) = connection?.split();
+
+        let clone_rt = runtime.clone();
         std::thread::spawn(move || {
             trace!("Starting msg dispatching loop");
-            Self::dispatch_incoming_messages(websocket_receiver, messages_tx, process_id);
+            clone_rt.block_on(Self::dispatch_incoming_messages(
+                recv,
+                messages_tx,
+                process_id,
+            ));
             trace!("Quit loop msg dispatching loop");
         });
 
         Ok(Self {
-            sender: Mutex::new(sender),
+            sender: Arc::new(Mutex::new(sender)),
+            runtime,
             process_id,
         })
     }
@@ -49,27 +67,27 @@ impl WebSocketConnection {
             "Shutting down WebSocket connection for Chrome {:?}",
             self.process_id
         );
-        if self.sender.lock().unwrap().shutdown_all().is_err() {
-            debug!(
-                "Couldn't shut down WS connection for Chrome {:?}",
-                self.process_id
-            );
-        }
+
+        self.runtime.block_on(async {
+            if self.sender.lock().await.close().await.is_err() {
+                debug!(
+                    "Couldn't shut down WS connection for Chrome {:?}",
+                    self.process_id
+                );
+            }
+        });
     }
 
-    fn dispatch_incoming_messages(
-        mut receiver: websocket::receiver::Reader<TcpStream>,
+    async fn dispatch_incoming_messages(
+        mut receiver: SplitStream<WebSocketStream<MaybeTlsStream<TcpStream>>>,
         messages_tx: mpsc::Sender<Message>,
         process_id: Option<u32>,
     ) {
-        for ws_message in receiver.incoming_messages() {
+        loop {
+            let ws_message = receiver.next().await;
             match ws_message {
-                Err(error) => match error {
-                    WebSocketError::NoDataAvailable => {
-                        debug!("WS Error Chrome #{:?}: {}", process_id, error);
-                        break;
-                    }
-                    WebSocketError::IoError(err) => {
+                Some(Err(error)) => match error {
+                    tungstenite::Error::Io(err) => {
                         debug!("WS IO Error for Chrome #{:?}: {}", process_id, err);
                         break;
                     }
@@ -78,8 +96,8 @@ impl WebSocketConnection {
                         process_id, error
                     ),
                 },
-                Ok(message) => {
-                    if let OwnedMessage::Text(message_string) = message {
+                Some(Ok(message)) => {
+                    if let tungstenite::Message::Text(message_string) = message {
                         if let Ok(message) = parse_raw_message(&message_string) {
                             if messages_tx.send(message).is_err() {
                                 break;
@@ -94,28 +112,23 @@ impl WebSocketConnection {
                         panic!("Got a weird message: {:?}", message);
                     }
                 }
+                // Keep waiting for an event in case nothing was fetched yet.
+                None => (),
             }
         }
 
         info!("Sending shutdown message to message handling loop");
         if messages_tx.send(Message::ConnectionShutdown).is_err() {
-            warn!("Couldn't send message to transport loop telling it to shut down");
+            warn!("Couldn't send message to transport loop telling it to shut down")
         }
     }
 
-    pub fn websocket_connection(ws_url: &Url) -> Result<Client<TcpStream>> {
-        let client = ClientBuilder::from_url(ws_url).connect_insecure()?;
-
-        debug!("Successfully connected to WebSocket: {}", ws_url);
-
-        Ok(client)
-    }
-
     pub fn send_message(&self, message_text: &str) -> Result<()> {
-        let message = websocket::Message::text(message_text);
-        let mut sender = self.sender.lock().unwrap();
-        sender.send_message(&message)?;
-        Ok(())
+        let message = tungstenite::Message::text(message_text);
+
+        Ok(self
+            .runtime
+            .block_on(async { self.sender.lock().await.send(message).await })?)
     }
 }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,12 @@ derive_builder = "0.12"
 directories = { version = "4.0", optional = true }
 log = "0.4"
 rand = "0.8"
+futures = "0.3.25"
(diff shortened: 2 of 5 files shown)


1c3t3a:
BTW, for me the current `master` doesn't build because the `env_logger` version is non-existent. The error:
```
error: failed to select a version for the requirement `env_logger = "^0.10"`
candidate versions found which didn't match: 0.9.3, 0.9.2, 0.9.1, ...
```

Billy-Sheppard:
Looks good! Thanks for your work on this - I am unable to test this atm but I'll merge it in in the next day or two, unless @mdrokz gets to it first 😆

src/browser/transport/web_socket_connection.rs
```diff
@@ -1,45 +1,63 @@
...
+
+        let connection: Result<_> = runtime.block_on(async {
+            let (stream, _) = connect_async(ws_url.to_string().into_client_request()?).await?;
 
+            debug!("Successfully connected to WebSocket: {}", ws_url);
+
+            Ok(stream)
+        });
+
+        let (sender, recv) = connection?.split();
```
mdrokz, line 45: does this have any performance or memory impact as we are using a async runtime and then blocking on the result

mdrokz:
> Currently this crate uses the `websocket` crate, that has a dependency on a hyper version that is subject to a security vulnerability. This change switches the implementation to the `tokio-tungstenite` websocket library, which exposes an async interface. In order to fit the async interface with the current sync interface, [async-bridging](https://tokio.rs/tokio/topics/bridging) is used.
> 
> This fixes #322. Related issues: #346, #341

and are all the test cases passing after using async bridging ? and has this been tested extensively ?

src/browser/transport/web_socket_connection.rs
1c3t3a: Apparently yes. This runtime object has some memory overhead since it contains schedulers, handles, etc. for a Tokio runtime. However, the execution speed shouldn't be too affected, but I also didn't run benchmarks. This runtime is created in the minimal invasive approach possible: by not spawning new threads but instead scheduling futures in the current thread. This "inlines" the web socket code whenever an event occurs. 

This is not perfect, but the only way to not rewrite the whole interface as async and at the same time mitigate the security vulnerability. I would assume that the memory and execution duration overhead is not notable at all to the end user, but again I did not run benchmarks.

1c3t3a:
> > Currently this crate uses the `websocket` crate, that has a dependency on a hyper version that is subject to a security vulnerability. This change switches the implementation to the `tokio-tungstenite` websocket library, which exposes an async interface. In order to fit the async interface with the current sync interface, [async-bridging](https://tokio.rs/tokio/topics/bridging) is used.
> > This fixes #322. Related issues: #346, #341
> 
> and are all the test cases passing after using async bridging ? and has this been tested extensively ?

The test suite ran through and I tested this with spawning a headless chrome process and attaching to it. I am not an expert in using this crate, so it might be better if you verify this. I tested your examples and replaced the new browser process with a connection to my existing process.

1c3t3a:
Force pushed to rewrite message handling loop.

src/browser/transport/web_socket_connection.rs
mdrokz: > Apparently yes. This runtime object has some memory overhead since it contains schedulers, handles, etc. for a Tokio runtime. However, the execution speed shouldn't be too affected, but I also didn't run benchmarks. This runtime is created in the minimal invasive approach possible: by not spawning new threads but instead scheduling futures in the current thread. This "inlines" the web socket code whenever an event occurs.
> 
> This is not perfect, but the only way to not rewrite the whole interface as async and at the same time mitigate the security vulnerability. I would assume that the memory and execution duration overhead is not notable at all to the end user, but again I did not run benchmarks.

Thank you for explaining all of this and yes this is a good enough solution for now, great job

mdrokz:
> > > Currently this crate uses the `websocket` crate, that has a dependency on a hyper version that is subject to a security vulnerability. This change switches the implementation to the `tokio-tungstenite` websocket library, which exposes an async interface. In order to fit the async interface with the current sync interface, [async-bridging](https://tokio.rs/tokio/topics/bridging) is used.
> > > This fixes #322. Related issues: #346, #341
> > 
> > 
> > and are all the test cases passing after using async bridging ? and has this been tested extensively ?
> 
> The test suite ran through and I tested this with spawning a headless chrome process and attaching to it. I am not an expert in using this crate, so it might be better if you verify this. I tested your examples and replaced the new browser process with a connection to my existing process.

Ok i will test this on some of my projects and see

Sytten:
The other approach would have been to rewrite the logic to avoid splitting the sender and receiver. They dont really need to be split AFAIK.

1c3t3a:
> The other approach would have been to rewrite the logic to avoid splitting the sender and receiver. They dont really need to be split AFAIK. Then you can use the sync tungstenite.

Yes, but that is much much slower and presents issues with the receiving loop. You would need to set a certain receive timeout, as calling `receive` blocks all send calls (and the other way around, though `send` calls are usually faster). This could lead to weird synchronization errors. But yes, it would be doable as well.

vvilhonen:
Would be transparent to the crate user and there's not much difference in the binary size:

**before:**
```
cargo bloat --example query_wikipedia --release --crates
    Finished release [optimized] target(s) in 0.06s
warning: the following packages contain code that will be rejected by a future version of Rust: traitobject v0.1.0
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 24`
    Analyzing target/release/examples/query_wikipedia

 File  .text     Size Crate
23.0%  66.7%   3.0MiB headless_chrome
 5.6%  16.2% 736.6KiB std
 1.4%   4.2% 191.1KiB regex_syntax
 1.4%   4.1% 186.9KiB regex
 0.6%   1.7%  77.6KiB aho_corasick
 0.3%   0.9%  40.6KiB url
 0.2%   0.7%  30.2KiB idna
 0.2%   0.5%  24.1KiB websocket
 0.2%   0.5%  23.4KiB hyper
 0.1%   0.4%  17.2KiB websocket_base
 0.1%   0.4%  17.1KiB serde_json
 0.1%   0.3%  12.7KiB memchr
 0.1%   0.2%  10.7KiB miniz_oxide
 0.1%   0.2%   8.5KiB rand_chacha
 0.0%   0.1%   6.3KiB rand_hc
 0.0%   0.1%   5.9KiB serde
 0.0%   0.1%   5.7KiB rand_os
 0.0%   0.1%   5.4KiB base64
 0.0%   0.1%   5.1KiB tempfile
 0.0%   0.1%   4.7KiB httparse
 0.2%   0.6%  28.3KiB And 19 more crates. Use -n N to show more.
34.5% 100.0%   4.4MiB .text section size, the file size is 12.9MiB

Note: numbers above are a result of guesswork. They are not 100% correct and never will be.
```

**after:**
```
cargo bloat --example query_wikipedia --release --crates
warning: unused manifest key: dependencies.tokio.no-default-features
    Finished release [optimized] target(s) in 0.08s
    Analyzing target/release/examples/query_wikipedia

 File  .text     Size Crate
22.7%  64.8%   3.0MiB headless_chrome
 5.9%  17.0% 815.5KiB std
 1.4%   4.0% 191.1KiB regex_syntax
 1.4%   3.9% 186.8KiB regex
 0.6%   1.6%  77.6KiB aho_corasick
 0.6%   1.6%  76.7KiB tokio
 0.3%   0.9%  43.9KiB tungstenite
 0.3%   0.9%  42.9KiB url
 0.2%   0.6%  30.7KiB idna
 0.2%   0.5%  22.5KiB http
 0.1%   0.4%  18.2KiB serde_json
 0.1%   0.3%  16.4KiB hashbrown
 0.1%   0.3%  12.7KiB memchr
 0.1%   0.2%  10.7KiB miniz_oxide
 0.1%   0.2%   8.5KiB rand_chacha
 0.0%   0.1%   5.9KiB serde
 0.0%   0.1%   5.3KiB sha1
 0.0%   0.1%   5.1KiB tempfile
 0.0%   0.1%   5.0KiB bytes
 0.0%   0.1%   4.9KiB httparse
 0.2%   0.5%  25.4KiB And 21 more crates. Use -n N to show more.
35.1% 100.0%   4.7MiB .text section size, the file size is 13.4MiB

Note: numbers above are a result of guesswork. They are not 100% correct and never will be.
```

Uskrai:
i've got a working implementation using sync tungstenite without pulling any async runtime here https://github.com/Uskrai/rust-headless-chrome/commit/92024b58c3e44d28112b439ca4759988e41cff44. it uses TcpStream::set_read_timeout to make the message handle loop doesn't block the mutex too long. i've used it on other project and it work pretty well. idk if  i should open another pr after seeing this one.

1c3t3a:
Any decisions on what way to go?

Billy-Sheppard:
@Uskrai - can you please open a PR for your fix? 

I'm not fully across the differences between the proposed changes here, and the ones in your commit, however if yours works and this one is still under consternation I think we can at least merge in yours while this is worked on more?

I'll let @mdrokz have the final say (as I am still working my way to understanding the inner workings of the crate).

Thanks to all of you for your work on this fix though, especially through the holidays (which might be causing some understandable delays 😄)

1c3t3a:
> @Uskrai - can you please open a PR for your fix?
> 
> I'm not fully across the differences between the proposed changes here, and the ones in your commit, however if yours works and this one is still under consternation I think we can at least merge in yours while this is worked on more?
> 
> I'll let @mdrokz have the final say (as I am still working my way to understanding the inner workings of the crate).
> 
> Thanks to all of you for your work on this fix though, especially through the holidays (which might be causing some understandable delays 😄)

This patch is also done from my side. It is rather the question of whether you want to go with the async runtime approach or the sync-but-not-splittable approach.

Billy-Sheppard:
From my perspective I would learn towards keeping it non-async for now. 

I think its reasonable to keep this code somewhat handy however because if someone wanted to move rust-headless-chrome to `async` (behind a feature perhaps) in the future then this would make more sense. I'll still await thoughts from mdrokz also.

mdrokz:
> From my perspective I would learn towards keeping it non-async for now.
> 
> I think its reasonable to keep this code somewhat handy however because if someone wanted to move rust-headless-chrome to `async` (behind a feature perhaps) in the future then this would make more sense. I'll still await thoughts from mdrokz also.

Im also leaning towards keeping it non async and PR #354 doesnt have too many downsides as this one in terms of memory and performance but i also agree with you to keep this code for the future in case we of an async API we can use a feature gate for async and sync API's

Billy-Sheppard:
I'm going to close out this request now since we've merged in the `sync` fix (#354).

PR #351, opened 2022-12-27
Fix docs and remove generated code

vvilhonen:
- Fix cargo doc warnings
- Remove generated code from version control
- Use updated auto_generate_cdp which generates the code to OUT_DIR

Should fix docs.rs build


Cargo.toml
```diff
@@ -40,14 +40,12 @@ png = { version = "0.17" }
...
+auto_generate_cdp = "0.4.0"
 
 [lib]
 name = "headless_chrome"
 path = "src/lib.rs"
 
 [features]
-default = ["auto_generate_cdp/native-tls"]
 fetch = ["ureq", "directories", "zip", "walkdir"]
-rustls = ["auto_generate_cdp/rustls"]
```
mdrokz, line 52: we are getting rid of this because of ureq right ? correct me if im wrong

vvilhonen: That's correct, it wasn't enabling anything else than the optional reqwest dependencies on auto_generate_cdp side

vvilhonen: auto_generate_cdp would need a new crates.io release before this compiles though, I can update the PR if you choose something else than 0.4.0 for that

mdrokz: > That's correct, it wasn't enabling anything else than the optional reqwest dependencies on auto_generate_cdp side

cool makes sense

mdrokz: > auto_generate_cdp would need a new crates.io release before this compiles though, I can update the PR if you choose something else than 0.4.0 for that

no 0.4.0 sounds good let me publish the release

mdrokz (approved):

PR #359, opened 2023-01-08
Fix usage of String::truncate

kawaemon:
I got a panic when passing `expression` with multi-byte characters to `browser::tab::Tab::evaluate`.
I found out the problem is wrong usage of `String::truncate`, so this PR replaces these usages to safer method.

e.g.
```rs
// this is fine.
String::from("abcde").truncate(1);
// this panics because character '★' is 2-bytes long, and we can't split it halfway.
String::from("★★★").truncate(1);
```

### DIFF
--- a/src/browser/mod.rs
+++ b/src/browser/mod.rs
@@ -371,9 +371,11 @@ impl Browser {
                                 }
                             }
                             _ => {
-                                let mut raw_event = format!("{:?}", event);
-                                raw_event.truncate(50);
-                                trace!("Unhandled event: {}", raw_event);
+                                let raw_event = format!("{:?}", event);
+                                trace!(
+                                    "Unhandled event: {}",
+                                    raw_event.chars().take(50).collect::<String>()
+                                );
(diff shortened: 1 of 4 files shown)


Billy-Sheppard:
Looking good, just needs a `cargo fmt` I think and it should be good to go.

kawaemon:
@Billy-Sheppard 
I ran `cargo fmt` and commited as [9d89c49](https://github.com/atroche/rust-headless-chrome/pull/359/commits/9d89c494098bafd33d7b492a674c805dcdffa214), but CI still says `cargo fmt` made diffs.
Looks like it ran on wrong commit? :sweat_smile: Idk why...

image

kawaemon:
I rebased it onto the latest commit of master branch. Could you run CI again? I believe it solves this problem.

kawaemon:
I think we're better to merge PRs when all workflows are passed.
This PR has a lot of unrelated diffs... This situation should be avoided.

PR #362, opened 2023-01-26
Fix to prevent panic on looping through empty vector

dogeatdog:
I didn't have access to create a PR through my desktop, so describing it like this. 
In browser/tab/element/mod.rs I've added a check around the loop (line 80) : 

`
```
if attrs_vec.len() > 0 {
            for i in (0..attrs_vec.len() - 1).step_by(2) {
                let k = attrs_vec[i].to_string();
                let v = attrs_vec[i+1].to_string();
                attrs.insert(k, v);
            }
        }
````

That will prevent a panic when there are no elements in the vector. (overflow in dev, index out of bounds in release.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "headless_chrome_fork"
-version = "1.0.0"
+version = "1.0.1"
 authors = ["Alistair Roche <alistair@sunburnt.country>"]
 edition = "2018"
 description = "Control Chrome programatically"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,12 +1,12 @@
 [package]
-name = "headless_chrome"
-version = "0.9.0"
+name = "headless_chrome_fork"
+version = "1.0.0"
 authors = ["Alistair Roche <alistair@sunburnt.country>"]
 edition = "2018"
 description = "Control Chrome programatically"
 license = "MIT"
-homepage = "https://github.com/atroche/rust-headless-chrome"
-repository = "https://github.com/atroche/rust-headless-chrome"
+homepage = "https://github.com/masc-it/rust-headless-chrome"
+repository = "https://github.com/masc-it/rust-headless-chrome"
 readme = "README.md"
 
 [dependencies]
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: Cargo.toml [hunk 1]
```diff
@@ -1,26 +1,26 @@
...
+version = "1.0.2"
 authors = ["Alistair Roche <alistair@sunburnt.country>"]
 edition = "2018"
 description = "Control Chrome programatically"
 license = "MIT"
-homepage = "https://github.com/atroche/rust-headless-chrome"
-repository = "https://github.com/atroche/rust-headless-chrome"
+homepage = "https://github.com/masc-it/rust-headless-chrome"
+repository = "https://github.com/masc-it/rust-headless-chrome"
 readme = "README.md"
```
mdrokz, line 10: can you please revert this ?

Cargo.toml
```diff
@@ -1,26 +1,26 @@
...
 
 [dependencies]
-websocket = { version = "0.23", default_features = false, features = ["sync"] }
+websocket = { version = "0.24", default_features = false, features = ["sync"] }
```
mdrokz, line 13: the latest master uses tungstenite can you use deps from master ?

README.md
```diff
@@ -1,7 +1,7 @@
 # Headless Chrome
 
 [![Build Status](https://travis-ci.com/atroche/rust-headless-chrome.svg?branch=master)](https://travis-ci.com/atroche/rust-headless-chrome)
-[![Crate](https://img.shields.io/crates/v/headless_chrome.svg)](https://crates.io/crates/headless_chrome)
+[![Crate](https://img.shields.io/crates/v/headless_chrome.svg)](https://crates.io/crates/headless_chrome_fork)
```
mdrokz, line 4: revert this as well

mdrokz (changes requested):
so far it looks good can you explain the storing document change ? and please revert the changes to `Cargo.toml` and `README.md`

mdrokz:
there are some conflicts as well can you resolve them properly ?

dogeatdog:
I goofed up. I was actually trying to create a PR to the forked repo https://github.com/masc-it/rust-headless-chrome, but somehow totally got lost in github. Hence the reason for the weird PR and the fact I just closed it right now. 

Sorry for the confusion !

PR #363, opened 2023-01-30
Fixed `clippy` lints + `docs.rs` fix

Billy-Sheppard:
Fixed a few clippy warnings (and bypassed some others).

Also added a `#[cfg(not(docsrs))]` which should help with the docs.rs issues.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,3 +1,4 @@
+#[cfg(not(docsrs))]
 use auto_generate_cdp::init;
 
 fn main() {
--- a/examples/local_storage.rs
+++ b/examples/local_storage.rs
@@ -18,7 +18,7 @@ fn main() -> Result<()> {
 
     let item: String = tab.get_storage("translationHash")?;
 
-    println!("{}", item);
+    println!("{item}");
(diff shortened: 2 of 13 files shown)


build.rs
```diff
@@ -1,3 +1,4 @@
+#[cfg(not(docsrs))]
 use auto_generate_cdp::init;
 
 fn main() {
```
mdrokz, line 4: The whole crate depends on `protocol.rs` in order to compile if we do this the `protocol.rs` file wont be generated

mdrokz (changes requested):

build.rs
Billy-Sheppard: Damn - afaik the only other option is to download a local version of the file and use that - I presume that will require repo changes in `auto_generate_cdp`

mdrokz: > Damn - afaik the only other option is to download a local version of the file and use that - I presume that will require repo changes in `auto_generate_cdp`

Let me see if i can include the file based on if its building inside docs.rs

PR #365, opened 2023-02-04
call get_document() in find_elements_by_xpath

masc-it:
**Problem**
If you run tab.find_elements_by_xpath(..), tab.wait_for_elements_by_xpath() as first commands in your file, they won't work.

**Solution**
Add a call to get_document() in find_elements_by_xpath(..) function, in the same way as find_elements(..) already does.


Billy-Sheppard:
Looks good! Thanks! Just run a `cargo fmt` and should be good to go.

masc-it:
> Looks good! Thanks! Just run a `cargo fmt` and should be good to go.

done 🚀

mdrokz (approved):

PR #368, opened 2023-02-14
Fix for `wait_for_initial_tab` bug

### LINKED ISSUE #367: Breaking : Latest Chrome breaks creating Browser object, steps to reproduce included.
Hi,

After updating Chrome on my Linux machine to 110.0.5481.77 (Official Build) (64-bit) the Browser object doesn't seem to be working anymore. 

Neither new with LaunchOptions or default seems to function, both return an error (due to Compat { error: Timeout }). Since literally nothing changed overnight, apart from Chrome being updated, I suspect an update is needed.

I've looked at the implementation of the  default_executable method and it doesn't seem to contain the current executable for chrome anymore : 

"google-chrome-stable",
        "chromium",
        "chromium-browser",
        "chrome",
        "chrome-browser",

I've tested my theory by ln google-chrome-stable to chrome, but unfortunately that doesn't remedy the issue :( 

Are you seeing the same issue or do you need more info to reproduce this error ?

Billy-Sheppard:
Fixes #367

@dogeatdog please see if this fixes your issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ use headless_chrome::protocol::cdp::Page;
 fn browse_wikipedia() -> Result<(), Box<dyn Error>> {
     let browser = Browser::default()?;
 
-    let tab = browser.wait_for_initial_tab()?;
+    let tab = browser.new_tab()?;
 
     /// Navigate to wikipedia
     tab.navigate_to("https://www.wikipedia.org")?;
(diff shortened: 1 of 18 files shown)


dogeatdog:
It does fix it. Thanks !

solidiquis:
no freaking wonder

PR #372, opened 2023-02-16
Updated Chrome revision and added Mac_Arm chrome installation

initprism:
1. fetch revision 1095492 is the same as the current [chrome revision of puppeteer](https://github.com/puppeteer/puppeteer/blob/b14628738010ab80960a1834c2db85bbe1223bac/packages/puppeteer-core/src/revisions.ts#L21).

2. Added a dependency to install the Chrome browser on Arm-based macOS.


Billy-Sheppard:
Looks good to me

mdrokz (approved):
Looks good as well ! thanks for your contribution

PR #374, opened 2023-02-16
Added fn to get latest chrome revision(requires fetch feature)

initprism:


mdrokz:
Hey @initprism thanks for the PR but can you explain the use of this method to fetch latest chrome revision?

initprism:
The `latest_chrome_revision` function returns the latest chrome revision in the platform environment.
This can be used by modifying the revision of `FecherOptions` if you want to use the latest Chrome version.


```rust
let revision =  headless_chrome::browser::latest_chrome_revision()?;
let fetcher_options = headless_chrome::FetcherOptions::default()
    .with_revision(revision);

let launch_options = headless_chrome::LaunchOptionsBuilder::default()
    .fetcher_options(fetcher_options)
    .build()?;
let browser = Browser::new(launch_options)?;
```

IMO, I don't think it's correct to always install the latest chrome revision when enable the fetch feature.
Also, the reason this function is not included in the `fetcher` module is that the lib does not export `fetcher` to pub. This will have to be reflected later to be moved to the `fetcher` module or written in the appropriate method.

masc-it:
I think this is really useful @initprism, maybe it can be re-designed in a way that extends the current .with_revision("some-version") function, rather than adding a new function, e.g. we can define a `Version` enum that can be passed to it:

```rust
enum Version {
    Specific(S),  // S: Into<String>
    Latest
}

.with_revision(Version::Specific("634997"))
or 
.with_revision(Version::Latest)
```

src/browser/mod.rs
```diff
@@ -493,6 +493,33 @@ pub fn default_executable() -> Result<std::path::PathBuf, String> {
...
+        url = format!("{}/Mac/LAST_CHANGE", url);
+    } else if cfg!(windows) {
+        url = format!("{}/Win_x64/LAST_CHANGE", url);
+    } else {
+        // This code will panic on unsupported platforms.
+        #[allow(unreachable_code)]
+        {
+            unimplemented!("This platform is not supported")
+        }
+    }
```
mdrokz, line 516: maybe it would be better to refactor this into a match statement ?

initprism: Refactoring to match statement looks good. And reflecting the definition suggested by @masc-it looks good.

initprism: I'm sorry I closed this PR by mistake. (;_;)
please check https://github.com/rust-headless-chrome/rust-headless-chrome/pull/380

PR #375, opened 2023-02-17
Add derive(Debug) and tracing logging

elpiel:
Small improvement to add derive(Debug) to 2 structs and adding tracing logging for the CLI options used to start the headless browser.

With the old version I had the issue of `print_to_pdf` always working sequentially even though I used it with futures and I was trying to make multiple PDF at the same time. Will test if this issue has persisted.


Billy-Sheppard:
Looks good to me.

mdrokz (approved):
Looks good as well ! thanks for your contribution

PR #377, opened 2023-02-18
Cache Attributes, Tag Name and Value on Element creation

masc-it:
Generally, we want to access an Element attributes and basic props when we create it, so why calling `get_attributes()` when we can already cache useful dom information, during the same Element instantation? I found out that it is literally a zero cost operation.

**Also**
- `get_attributes()` calls `get_description()` which has a depth of 100
- From my experiments, on long and complex pages it saves a lot of overhead (even couple of seconds per page)

_Open point for another time_
- Why do `el.get_description()` and `tab.describe_node()` have a depth of 100? Maybe we can parametrize it

### DIFF
--- a/src/browser/tab/element/mod.rs
+++ b/src/browser/tab/element/mod.rs
@@ -31,6 +31,9 @@ pub struct Element<'a> {
     pub backend_node_id: DOM::NodeId,
     pub node_id: DOM::NodeId,
     pub parent: &'a super::Tab,
+    pub attributes: Option<Vec<String>>,
+    pub tag_name: String,
+    pub value: String,
 }
 
 impl<'a> Debug for Element<'a> {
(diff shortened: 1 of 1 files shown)


Billy-Sheppard:
This seems sensible, thanks!

As for your last point I don't have any direct familiarity with those methods - but I don't see any reason why they can't take a parameter - perhaps its worth adding this functionality as a new method as not to break API - something like `get_description_with_depth(&self, depth: Option<usize>) -> Result<DOM::Node> {}`

mdrokz (approved):

mdrokz:
> This seems sensible, thanks!
> 
> As for your last point I don't have any direct familiarity with those methods - but I don't see any reason why they can't take a parameter - perhaps its worth adding this functionality as a new method as not to break API - something like `get_description_with_depth(&self, depth: Option<usize>) -> Result<DOM::Node> {}`

This makes sense as well to ensure API compatibility otherwise everything looks good. @Billy-Sheppard you can merge this if you want

PR #386, opened 2023-03-25
tungstenite connect_with_config use max frame size and max message size

### LINKED ISSUE #385: Print to PDF fail when dealing with large pictures
Print to PDF will fail when dealing with large pictures, here's the error message:
```bash
thread '<unnamed>' panicked at 'Unhandled WebSocket error for Chrome #Some(96226): 
Capacity(MessageTooLong { size: 47175864, max_size: 16777216 })', 
.../src/browser/transport/web_socket_connection.rs:100:25
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error: Unable to make method calls because underlying connection is closed
```

It looks like the limit is caused by the protocol (`usize::max_value()`), https://github.com/snapview/tungstenite-rs/blob/edb237754084bfb18a301ac8a3a42612af64ae19/src/protocol/message.rs#L138-L151 It would be great if you can resolve this.

Here's the code for reproduction.

```rust
use std::fs;
use anyhow::Result;
use headless_chrome::{types::PrintToPdfOptions, Browser, LaunchOptionsBuilder};

fn main() -> Result<()> {
    let browser = Browser::new(LaunchOptionsBuilder::default().build()?)?;
    let tab = browser.new_tab()?;
    let pdf_options: Option<PrintToPdfOptions> = None;
    let local_pdf = tab
        .navigate_to("https://upload.wikimedia.org/wikipedia/commons/3/3e/Irises-Vincent_van_Gogh.jpg")?
        .wait_until_navigated()?
        .print_to_pdf(pdf_options)?;
    fs::write("Irises.pdf", local_pdf)?;

    Ok(())
}
```

HollowMan6:
Resolve #385 

https://docs.rs/tungstenite/latest/tungstenite/client/fn.connect_with_config.html
https://docs.rs/tungstenite/latest/tungstenite/protocol/struct.WebSocketConfig.html

### DIFF
--- a/src/browser/transport/web_socket_connection.rs
+++ b/src/browser/transport/web_socket_connection.rs
@@ -6,6 +6,7 @@ use std::sync::Mutex;
 use anyhow::Result;
 use log::{debug, info, trace, warn};
 use tungstenite::http::Response;
+use tungstenite::protocol::WebSocketConfig;
 use tungstenite::stream::MaybeTlsStream;
 use url::Url;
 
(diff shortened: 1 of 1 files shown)


mdrokz (approved):
Looks good! thanks for your contribution

PR #391, opened 2023-04-29
Make `--enable-logging` not hard coded

HollowMan6:
Resolve https://github.com/rust-headless-chrome/rust-headless-chrome/issues/371 according to the idea from @vootooguru and verification from @ameleumit

I add `enable_logging` to LaunchOptions and make it disabled by default.


mdrokz (approved):
Thanks for your contribution!

PR #395, opened 2023-06-12
Make --disable-gpu not hard coded

sashka:
Default is `false`, so this commit doesn't break current use cases.


mdrokz (approved):
Thanks for your contribution!

Issue #364, opened 2023-02-02
Crash at get_midpoint


makorne:
Hi!
Thank you for great crate!

After several successful click at the same button, something changed on page and:

`
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /home/den/.cargo/git/checkouts/rust-headless-chrome-fd32a97e7a7fa8b4/60b775b/src/browser/tab/element/mod.rs:510:51
`

 `let raw_quad = quad.quads.first().unwrap(); ` << this unwrap

Buttons div changed to `<div style="display: none;">`

masc-it:
Hi @makorne, are you using CSS or XPATH selector to find such button?

makorne:
Hi @masc-it 

```
while let Ok(button) = tab.wait_for_element(".button1"){ 
    match button.click() {
     ...
    }
}
```

Ratipoka:
I had the same issue trying to click on a button
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', C:\Users\me\.cargo\registry\src\github.com-1ecc6299db9ec823\headless_chrome-1.0.5\src\browser\tab\element\mod.rs:510:51
```
Looking in the code, the line @makorne pointed : 
`let raw_quad = quad.quads.first().unwrap();`
suggests that the quad array returned by the `GetContentQuad` method was empty
Looking in some [DOM docs](https://vanilla.aslushnikov.com/?DOM.getContentQuads) (and asking chatgpt) it could be that this happens if the element we are trying to click is not visible.

I've checked with `headless(false)` what could be blocking the view of the button and it turns out that it was hidden behind a cookie popup.

adding 
```rust
tab.wait_for_xpath(r#"//*[@id="bnp_btn_reject"]"#)?
    .click()?;
```
fixed the issue for me.

Maybe we could change the code to something like this : 
```rust
let raw_quad = quad.quads.first().except("trying to get the midpoint of an element which is not visible");
```
to get a more helpfull error than
```
'called `Option::unwrap()` on a `None` value'
```

PR #396, opened 2023-06-18
better error for click on element which is not visible

Ratipoka:
This is in response to [#364](https://github.com/rust-headless-chrome/rust-headless-chrome/issues/364)
to get a better error in the case we try to click on an element which is hidden / not visible.


mdrokz (approved):
Thanks for your contribution!

PR #403, opened 2023-07-24
Add `press_key_with_modifiers`

nickjhughes:
Identical to `press_key`, but with an optional u32 bitfield which can be constructed from a slice of `ModifierKey` elements to specify which modifiers could should be held down (forwarded directly to the `modifiers` parameter of the [`Input.dispatchKeyEvent`](https://chromedevtools.github.io/devtools-protocol/tot/Input/#method-dispatchKeyEvent).

### DIFF
--- a/src/browser/tab/mod.rs
+++ b/src/browser/tab/mod.rs
@@ -62,6 +62,14 @@ pub mod element;
 mod keys;
 pub mod point;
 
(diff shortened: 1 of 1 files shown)


mdrokz (approved):
Thanks for your contribution!

PR #405, opened 2023-07-31
Disable automatic tab creation on launching chrome

### LINKED ISSUE #392: How to avoid opening a tab when connecting to existing browser
When I connect to an already open browser I always get a new tab.

`Browser::connect(debug_ws_url)`

Is this expected? Can I avoid it?

dimchee:
It is useful to launch chrome directly with `<url>` (or `--app=<url>`) as argument, if you use headful browser.
Not opening empty tab is essential for [simple neovim plugin](https://github.com/dimchee/prochrome.nvim) I am developing. 
It would be nice too if `--app=<url>` option was part of `LaunchOptions`.
I have seen that others had same issue as me, so this fixes #392.

### DIFF
--- a/src/browser/mod.rs
+++ b/src/browser/mod.rs
@@ -164,11 +164,7 @@ impl Browser {
         // so we get events like 'targetCreated' and 'targetDestroyed'
         trace!("Calling set discover");
         browser.call_method(SetDiscoverTargets { discover: true })?;
-
-        let tab = browser.new_tab()?;
-
-        tab.call_method(DOM::Enable(None))?;
-        tab.call_method(CSS::Enable(None))?;
+        
         Ok(browser)
     }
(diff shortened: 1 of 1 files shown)


Billy-Sheppard:
This seems reasonable to me. What do you think @mdrokz?

mdrokz:
> This seems reasonable to me. What do you think @mdrokz?

Hey didnt get time to review this but this comment https://github.com/rust-headless-chrome/rust-headless-chrome/issues/392#issuecomment-1685795373 seems to be concerning since it might be breaking `get_tabs` method can you check this and let me know @Billy-Sheppard thanks ?

Billy-Sheppard:
Might have to be behind a flag, I can look at it in-environment soon.

PR #406, opened 2023-08-01
Fixes wrong error message when running as root in sandbox mode

fcoury:
When running as root the error reported is misleading:

```
There are no available ports between 8000 and 9000 for debugging
```

This commit fixes this and reports the proper (new) error `RunningAsRootWithoutNoSandbox`:

```
You need to set the sandbox(false) option when running as root
```

### DIFF
--- a/src/browser/process.rs
+++ b/src/browser/process.rs
@@ -273,7 +273,7 @@ impl Process {
             attempts += 1;
         }
 
-        let mut child = process.0.borrow_mut();
+        let child = process.0.borrow_mut();
         child.stderr = None;
 
         Ok(Self {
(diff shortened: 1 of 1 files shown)


Billy-Sheppard:
Looks good, can you please make sure it passes `cargo check` and then you're good to go! Thanks!

fcoury:
@Billy-Sheppard done even though the cargo check complaint was not caused by my changes and happens in `main` as well.

Thanks!

Billy-Sheppard:
Ah! My mistake, thank you - I'm currently working on a fix for some of these issues. Thanks for your PR!

PR #409, opened 2023-08-09
Fix detecting msedge.exe in Windows when Chrome is not found through registry

CrendKing:
I don't have Chrome installed, but Edge exists by default. When I do `println!("{}", headless_chrome::browser::default_executable());`, it tells me "Could not auto detect a chrome executable".

The related msedge.exe detection code seems wrong, because 1) it doesn't use `path` from the enclosing if let, and 2) msedge should be only needed if chrome is not detected, but the current code only does it otherwise.


Billy-Sheppard:
This looks good! Thanks.

PR #410, opened 2023-08-09
Propagate error of unable to find midpoint upward instead of panicking

CrendKing:
It seems sometimes after finding an element, if immediately calling something like `.click()`, we get a panic at unwrapping a `None`. The related code seems to assume this should never happen. Maybe it's a better idea to propagate the error upward.


mdrokz (approved):
Thanks for your contribution!

PR #411, opened 2023-08-10
Add get_attribute_value() for Element

CrendKing:
Seemingly a quintessential function for an Element object, and there is indeed [people asking for it](https://github.com/rust-headless-chrome/rust-headless-chrome/issues/308), and [chromiumoxide implements it](https://github.com/mattsse/chromiumoxide/blob/939a12664ac6fc85fde92112f80bcafdb3e26421/src/element.rs#L334).


mdrokz (approved):
Thanks for the contribution!

PR #469, opened 2024-03-08
chore(msrv): set to 1.70

0x61nas:
It will be helpful for the down stream crates to know what the minimum supported Rust version (MSRV) for their dependencies so they can decide how to bump their version and when, because it's considered as a breaking change

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ name = "headless_chrome"
 version = "1.0.9"
 authors = ["Alistair Roche <alistair@sunburnt.country>"]
 edition = "2021"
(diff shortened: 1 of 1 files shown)


kwinz:
As a side note: can the maintainers please approve the workflow?
Some pull request changed the workflow file, but the change to the config wasn't approved yet.
Apparently the integration tests are being run, but the results haven't shown up because of the missing approval.
And I can see that the tests are [failing](https://github.com/rust-headless-chrome/rust-headless-chrome/actions/runs/8045112886/job/21969817039#step:5:213) so bugs have already made it into the release!

PR #476, opened 2024-04-08
Support linux aarch64 Chrome downloads

pruthvikar:
Currently, any arch of Linux will download an x86_64 installation of Chrome. This PR handles downloading arm64 installations.


pruthvikar:
Turns out, these builds don't actually exist!

PR #492, opened 2024-07-11
Bugfix: iterator instead of string was printed

TonalidadeHidrica:
The `trace!` output of `Tab::call_method` was trying to print the first 70 characters of the result of the method call, but it was printing an iterator instead.  As a result, it print much more long log even than the original string despite the intention.  I fixed the bug.


Billy-Sheppard:
Thanks!

PR #493, opened 2024-07-11
Replace deprecated `default_features` with `default-features`

TonalidadeHidrica:
I compiled the crate and found a warning complaining about an obsolete use of `deafult_features`:

```
warning: `default_features` is deprecated in favor of `default-features` and will not work in the 2024 edition
(in the `chrono` dependency)
warning: `default_features` is deprecated in favor of `default-features` and will not work in the 2024 edition
(in the `jpeg-decoder` dependency)
```

so I fixed it.


Billy-Sheppard:
Thanks!

PR #539, opened 2025-02-19
feat(cdp): update protocol and other dependencies

nileshmali:
Sorry for the duplicate PR, just wanted a quick resolution.

This PR contains protocol + dependency updates.


Billy-Sheppard:
Looks good - just one quick question - why has the version been bumped so much?

PR #566, opened 2026-06-09
Rustls tls webpki roots

jeltou:
### Background

`headless_chrome` currently relies on Tungstenite's default TLS configuration when establishing WebSocket connections to the Chrome DevTools endpoint.

In environments that use an internal PKI or private certificate authority, the DevTools WebSocket endpoint may be served using certificates that are not trusted by the system or bundled root stores. In such cases, establishing a connection fails with errors similar to:

```text
IO error: invalid peer certificate: UnknownIssuer
```

This makes it difficult to use `headless_chrome` in enterprise environments where internal services are exposed through TLS using private root certificates.

### What changes

This PR adds an alternative connection path that allows callers to provide additional root certificates when creating a WebSocket connection.

The implementation:

* Adds support for supplying custom root certificates as PEM or DER encoded bytes.
* Creates a dedicated Rustls `RootCertStore`.
* Loads the provided certificate(s) into the trust store.
* Establishes the WebSocket connection using a custom Rustls connector.
* Preserves the existing connection behavior when no custom certificate is provided.

The default behavior remains unchanged for existing users.

### Motivation

This change enables secure connections to Chrome DevTools endpoints that are exposed behind:

* internal PKI infrastructures
* enterprise reverse proxies
* self-hosted browser clusters
* private certificate authorities

without requiring global trust-store modifications or disabling certificate validation.

### Compatibility

The feature is completely opt-in.

If no custom root certificate is provided, the existing connection mechanism is used and behavior remains unchanged.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,8 +28,9 @@ tungstenite = "0.28"
 url = "2.5"
 which = "8.0"
 zip = { version = "6.0.0", optional = true }
-rustls = { version = "0.23" }
-rustls-pemfile = "2.2.0"
+rustls = { version = "0.23", optional = true }
+rustls-pemfile = { version = "2.2.0", optional = true }
+webpki-roots = { version = "1.0.7", optional = true }
 
 [target.'cfg(windows)'.dependencies]
 winreg = "0.55.0"
@@ -55,5 +56,5 @@ fetch = ["ureq", "directories", "zip", "walkdir"]
 nightly = []
 offline = ["auto_generate_cdp/offline"]
 rustls-tls-native-roots = ["tungstenite/rustls", "tungstenite/rustls-tls-native-roots"]
-rustls-tls-webpki-roots = ["tungstenite/rustls", "tungstenite/rustls-tls-webpki-roots", "rustls/ring"]
+rustls-tls-webpki-roots = ["tungstenite/rustls", "tungstenite/rustls-tls-webpki-roots", "rustls", "rustls-pemfile", "webpki-roots",  "rustls/ring"]
 native-tls = ["tungstenite/native-tls"]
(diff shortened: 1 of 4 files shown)


src/browser/transport/web_socket_connection.rs
```diff
@@ -118,74 +188,113 @@ impl WebSocketConnection {
...
+
+            let port = ws_url
+                .port_or_known_default()
+                .ok_or_else(|| anyhow::anyhow!("missing websocket port: {ws_url}"))?;
 
-        debug!("Successfully connected to WebSocket: {ws_url}");
+            let tcp = TcpStream::connect((host, port))?;
 
-        Ok(client)
+            let mut roots = rustls::RootCertStore::empty();
```
Billy-Sheppard, line 257: When a custom root certificate is provided, this creates an empty cert store and only adds the user provided cert. Is that intentional? Users may expect the custom certificate to be added alongside the default system/WebPKI roots rather than replacing them entirely? Not sure

Cargo.toml
```diff
@@ -28,6 +28,8 @@ tungstenite = "0.28"
 url = "2.5"
 which = "8.0"
 zip = { version = "6.0.0", optional = true }
+rustls = { version = "0.23" }
```
Billy-Sheppard: The new dependencies appear to be needed only for the new feature. Should they be `optional = true` and tied to the relevant rustls feature, to save extra compiling?

src/browser/transport/web_socket_connection.rs
jeltou: Good point. My original use case was connecting to an endpoint secured by a private CA, so I started with an empty trust store.I can update the implementation to add the provided certificates on top of the existing WebPKI roots instead of replacing them

Cargo.toml
jeltou: i adjusted the Cargo.toml. The dependecies are tied to the feature and only compile if the feature is activated

